PR #3, opened 2022-01-10
Hubspot Forms integration.

AndyButland:
This PR contains the initial commit for an Umbraco/Hubspot Forms integration, based on the code provided with some small updates for aligning naming, namepsaces and patterns for consistency with other integrations.

To finish off and release we need the following:

- [x] Test existing integration with Hubspot using API key authentication
- [x] Update use of `umb-overlay` (see comment from Bjarne)
- [x] Add OAuth option for authenticating re-using code/flow from the [Umbraco Forms/Hubspot integration](https://github.com/umbraco/Umbraco.Forms.Integrations/tree/main/src/Umbraco.Forms.Integrations.Crm.Hubspot)
- [x] Unit tests (e.g. for `FormsController`)
- [x] Error handling for failed API requests (in `FormsController`)
- [x] Build pipeline

Following that we can release to NuGet and our and blog about the new integration.


src/Umbraco.Cms.Integrations.Crm.Hubspot/App_Plugins/UmbracoCms.Integrations/Crm/Hubspot/views/formpicker.html
```diff
@@ -0,0 +1,32 @@
...
+
+    <div ng-if="vm.error" class="alert alert-warning">
+        {{ vm.error }}
+    </div>
+
+    <umb-overlay ng-if="vm.hubspotFormPickerOverlay.show"
+                 model="vm.hubspotFormPickerOverlay"
+                 view="vm.hubspotFormPickerOverlay.view"
+                 position="right">
+    </umb-overlay>
```
bjarnef, line 31: @AndyButland while `<umb-overlay>` still work, we should instead use editorService or overlayService.
I think we only have a single reference left in core using `<umb-overlay>`.

acoumb:
Integration points completed.

acoumb:
Items reviewed and updated.

AndyButland:
Thanks for the updates.  Got a few more points following review this morning.

- Still thinking we have too much in configuration that we don't want or expect the user to change - basically all the OAuth options.  I don't see we would want people to change these and hence rather than asking them to add them, they are better hard-coded in the package (like we did for the Umbraco Forms one).  So I think the only thing you should add in configuration is the API key, and if that's not there, we assume OAuth using the client Id, scopes etc. that are hard-coded as constants in the package.

- Can see in `HubspotForm.cshtml` that the `region` setting is still set to a fixed value of `eu1`.  That's OK if there's truly only one value of these, but I was assuming there may be others, e.g. `us1.  So we need to make that a property on `HubspotFormViewModel` and populate it from somewhere.  Ideally the API request we make to HubSpot to get the forms where I assume it must be available in the response.  If not, would have to be a config value.

- I'm getting an "error" response when trying to connect with OAuth.  Could be some settings that need updating though so will look at that again later.

- Just a small point on the UX.  Think we can be a little bit more explanatory in the datatype settings to help out the developer.  Rather than the big green alert, think we can tone down the design a bit and explain more.  So state one of:
   - "An API key is configured and will be used to connect to your HubSpot account."
   - "No API key is configured. To connect to your HubSpot account using OAuth click 'Connect', select your account and agree to the permissions."
   - "OAuth is configured and an access token is available to connect to your HubSpot account.  To revoke, click 'Revoke'".

acoumb:
Integration updated.

src/Umbraco.Cms.Integrations.Crm.Hubspot/Controllers/FormsController.cs
```diff
@@ -38,6 +38,7 @@ public class FormsController : UmbracoAuthorizedApiController
         private const string OAuthScopes = "oauth forms"; // oauth forms
         private const string OAuthProxyBaseUrl = "https://localhost:44364/"; //"https://hubspot-forms-auth.umbraco.com/"; // for local testing: https://localhost:44364
         private const string OAuthProxyEndpoint = "{0}oauth/v1/token";
+        private const string HubspotServiceName = "HubspotForms";
```
bjarnef, line 41: Should probably be `HubSpot` instead of `Hubspot` :)

PR #6, opened 2022-02-22
V9 multi-target updates.

acoumb:
Property Editor updates for multi-targeting.
Fix unit tests.


AndyButland (approved):
This all looks fine to me, good job.

Only thing I'd add - which can be a separate task after you've merged this PR - would be to add a V9 test site to the solution.  There's one in the repo already (though not in the solution), so you can use that - but it should be renamed to be generic (i.e. not just for CommerceTools), and as it's based on an V9 RC you may be better deleting it and creating a new one.

Then when you have it set it up with a reference to the HubSpot project it can be tested.

One thing you may be missing is a targets file in the multi-targeted .cspoj file (which is needed to make sure the client-side assets get copied into the NuGet package).  See [here](https://umbraco.com/blog/multi-targeted-package-migration/), starting from _"We also need to add to the project file two further elements..."_, in [this forum post](https://our.umbraco.com/forum/using-umbraco-and-getting-started/108129-how-to-create-an-umbraco-9-plugin-nuget-package#comment-335921) and in action in this package ([here](https://github.com/umbraco/UI-Examples/blob/master/src/Our.Umbraco.UiExamples/Our.Umbraco.UiExamples.csproj#L45) and [here](https://github.com/umbraco/UI-Examples/blob/master/src/Our.Umbraco.UiExamples/build/Our.Umbraco.UiExamples.targets)).

PR #9, opened 2022-03-08
Fix rendering component for Umbraco 9; pass region parameter to rende…

acoumb:
…ring component using the view model.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Crm.Hubspot/App_Plugins/UmbracoCms.Integrations/Crm/Hubspot/Render/HubspotForm.cshtml
+++ b/src/Umbraco.Cms.Integrations.Crm.Hubspot/App_Plugins/UmbracoCms.Integrations/Crm/Hubspot/Render/HubspotForm.cshtml
@@ -1,10 +1,6 @@
 ﻿@inherits Umbraco.Web.Mvc.UmbracoViewPage<Umbraco.Cms.Integrations.Crm.Hubspot.Models.ViewModels.HubspotFormViewModel>
 
-@{
-    var hbsptScriptPath = string.IsNullOrEmpty(Model.Region) ? "//js.hsforms.net/forms/shell.js" : "//js-eu1.hsforms.net/forms/shell.js";
-}
-
-<script charset="utf-8" type="text/javascript" src="@hbsptScriptPath"></script>
+<script charset="utf-8" type="text/javascript" src="@Model.ScriptPath"></script>
 <script>
     hbspt.forms.create({
         region: "@Model.Region",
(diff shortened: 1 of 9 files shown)


src/Umbraco.Cms.Integrations.Crm.Hubspot/App_Plugins/UmbracoCms.Integrations/Crm/Hubspot/Render/HubspotForm.cshtml
```diff
@@ -1,31 +1,14 @@
...
-            portalId: "@Model.PortalId",
-            formId: "@Model.Id"
-        });
-    </script>
-
-}
\ No newline at end of file
+<script>
+    hbspt.forms.create({
+        region: "@Model.Region",
```
AndyButland, line 10: Looks like you are now applying the `region` setting irrespective of if we have a value now.  If that works as expected then it's fine, just noting in case you did this inadvertently.

src/Umbraco.Cms.Integrations.Crm.Hubspot/App_Plugins/UmbracoCms.Integrations/Crm/Hubspot/Render/HubspotForm.cshtml
```diff
@@ -1,31 +1,14 @@
-﻿@using System.Configuration
-@inherits Umbraco.Web.Mvc.UmbracoViewPage<Umbraco.Cms.Integrations.Crm.Hubspot.Models.ViewModels.HubspotFormViewModel>
+﻿@inherits Umbraco.Web.Mvc.UmbracoViewPage<Umbraco.Cms.Integrations.Crm.Hubspot.Models.ViewModels.HubspotFormViewModel>
 
 @{
-    var region = ConfigurationManager.AppSettings["Umbraco.Cms.Integrations.Crm.Hubspot.Region"];
-
-    var hbsptScriptPath = string.IsNullOrEmpty(region) ? "//js.hsforms.net/forms/shell.js" : "//js-eu1.hsforms.net/forms/shell.js";
+    var hbsptScriptPath = string.IsNullOrEmpty(Model.Region) ? "//js.hsforms.net/forms/shell.js" : "//js-eu1.hsforms.net/forms/shell.js";
```
AndyButland, line 4: Should `eu1` here in the script's host name actually come from the value of `Model.Region`?  In case there is more than one region?

src/Umbraco.Cms.Integrations.Crm.Hubspot/Umbraco.Cms.Integrations.Crm.Hubspot.csproj
```diff
@@ -42,6 +43,10 @@
 		</Content>
 	</ItemGroup>
 
+	<ItemGroup>
```
AndyButland, line 46: Don't think you need this in the csproj... you should check, but I suspect it was added by Visual Studio unnecessarily.

AndyButland (approved):
Looks fine - just left a few things to check.

PR #13, opened 2022-04-04
Zapier V9 multi-target

acoumb:


src/Umbraco.Cms.Integrations.Automation.Zapier/App_Plugins/UmbracoCms.Integrations/Automation/Zapier/lang/en-US.xml
```diff
@@ -2,5 +2,7 @@
 <language>
 	<area alias="dashboardTabs">
 		<key alias="zapierDashboard">Zapier Integrations</key>
+		<key alias="promptDeleteTitle">Delete Registered Webhook</key>
```
AndyButland, line 5: Possibly should use something here to be sure we don't clash with another package.  I.e. either rename the area to be `zapierDashboard` or rename the key aliases to be prefixed with `zapier...`.

src/Umbraco.Cms.Integrations.Automation.Zapier/Components/NewContentPublishedComponent.cs
```diff
@@ -2,18 +2,66 @@
 using System.Collections.Generic;
 using System.Net.Http;
 using System.Threading.Tasks;
+
```
AndyButland, line 5: I think this file could be amended a bit, as it's a bit difficult to follow given it's pretty much the whole class that changes, and the name of the class itself varies between the two versions.

Would suggest having two classes and files - and just fully include/exclude the whole file based on the compiler directive, like [this one](https://github.com/umbraco/UI-Examples/blob/master/src/Our.Umbraco.UiExamples/Migrations/UmbracoAppStartingHandler.cs).

src/Umbraco.Cms.Integrations.Automation.Zapier/Components/ZapContentConfigComponent.cs
```diff
@@ -1,15 +1,67 @@
...
+            _scopeProvider = scopeProvider;
+
+            _keyValueService = keyValueService;
+
+            _runtimeState = runtimeState;
+        }
+
+        public void Handle(UmbracoApplicationStartingNotification notification)
+        {
+            if(_runtimeState.Level < RuntimeLevel.Run) return;
```
AndyButland, line 53: Same comment applies to this class with regard it maybe being better as two files.

You can also look to move the migration plan out into it's own class, which then needs a smaller set of compiler directive include/excludes (see [this example](https://github.com/umbraco/UI-Examples/blob/master/src/Our.Umbraco.UiExamples/Migrations/UiExamplesMigrationPlan.cs)).

src/Umbraco.Cms.Integrations.Automation.Zapier/Migrations/ZapContentConfigTable.cs
```diff
@@ -1,30 +1,62 @@
...
+        public string MigrationLoggingMessage = $"Running migration {Constants.MigrationPlanName}";
+
+        public string DbTableExistsMessage =
+            $"The database table {Constants.ZapContentConfigTable} already exists, skipping";
+
         public ZapContentConfigTable(IMigrationContext context) : base(context)
         {
         }
 
+#if NETCOREAPP
```
AndyButland, line 27: You perhaps only need the compiler directive around the method signature and the logger, which means you don't need to repeat the actual migration logic itself.

src/Umbraco.Cms.Integrations.Automation.Zapier/App_Plugins/UmbracoCms.Integrations/Automation/Zapier/lang/en-US.xml
acoumb: Updated alias and keys.

acoumb:
Updated and tidied up components.

AndyButland:
Only thing I can still see is the migration plan, that you are currently defining it in two places.  It's not complicated, so not a big deal, but perhaps it'll get more involved in the future.  I think you should be able to define it in it's own class, like [this one](https://github.com/umbraco/UI-Examples/blob/master/src/Our.Umbraco.UiExamples/Migrations/UmbracoAppStartingHandler.cs).

acoumb:
Migration plan moved to an unique object.

PR #14, opened 2022-04-12
Subscription and Polling endpoints

acoumb:
To support validation requirements for app publish on Zapier marketplace, additional features have been added to the integration:

1. _IsEnabled_ flag on the content configuration to toggle whether the trigger hook should be called when the specific content gets published. By default, the marker will be _true_ when entities are registered in the _Zapier Integrations_ dashboard and managed through the _Subscription_ endpoints.
2. _Subscription_ endpoint - Zapier refers to the REST webhooks as static triggers, and requires endpoints for _subscribe/unsubscribe_ . The _UpdatePreferences_ action handles both use cases: it expects a request body containing the hook URL and a Boolean for enabled, and validates first each request using the header keys: _X-USERNAME_ and _X-PASSWORD_.
3. _Polling_ endpoint - fallback action required by Zapier to ensure that if a trigger is rendered unable, the Zapier app can pull the published content.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Automation.Zapier/Controllers/PollingController.cs
+++ b/src/Umbraco.Cms.Integrations.Automation.Zapier/Controllers/PollingController.cs
@@ -1,0 +1,104 @@
+﻿using System.Collections.Generic;
+
+using System.Linq;
+
+using Umbraco.Cms.Integrations.Automation.Zapier.Configuration;
+using Umbraco.Cms.Integrations.Automation.Zapier.Services;
+
+#if NETCOREAPP
+using Microsoft.Extensions.Options;
+
+using Umbraco.Cms.Web.Common.Controllers;
+using Umbraco.Cms.Core.Services;
+#else
+using System.Configuration;
+
+using Umbraco.Web.WebApi;
+using Umbraco.Core.Services;
+#endif
+
+namespace Umbraco.Cms.Integrations.Automation.Zapier.Controllers
+{
+    public class PollingController : UmbracoApiController
+    {
+        private readonly ZapierSettings Options;
+
+        private IContentService _contentService;
+
+        private readonly IUserValidationService _userValidationService;
+
+#if NETCOREAPP
+        public PollingController(IOptions<ZapierSettings> options, IContentService contentService, IUserValidationService userValidationService)
+#else
+        public PollingController(IContentService contentService, IUserValidationService userValidationService)
+#endif
+        {
+#if NETCOREAPP
+            Options = options.Value;
+#else
+            Options = new ZapierSettings(ConfigurationManager.AppSettings);
+#endif
+
+            _contentService = contentService;
+
+            _userValidationService = userValidationService;
+        }
+
+        public List<Dictionary<string, string>> GetPublishedContent()
+        {
+            var publishedContent = new List<Dictionary<string, string>>();
+
+            string username = string.Empty;
+            string password = string.Empty;
+
+#if NETCOREAPP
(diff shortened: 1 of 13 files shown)


src/Umbraco.Cms.Integrations.Automation.Zapier/Migrations/ZapContentConfigTable.cs
```diff
@@ -67,6 +67,9 @@ public class ZapContentConfig
 
             [Column("WebHookUrl")] 
             public string WebHookUrl { get; set; }
```
AndyButland, line 69: We should make sure `WebHookUrl` has a unique index set on it (as we're now using it for look-ups).
`ContentTypeName` should have one as well for similar reasons.

An example [here](https://github.com/umbraco/Forms/blob/v10/dev/src/Umbraco.Forms.Core/Persistence/Dtos/UserGroupFormSecurity.cs#L30) - though this isn't a unique index, but I think the attribute used here has options for that.

src/Umbraco.Cms.Integrations.Automation.Zapier/Services/UserValidationService.cs
```diff
@@ -0,0 +1,54 @@
...
+using Umbraco.Cms.Core.Security;
+using Umbraco.Cms.Core.Services;
+#else
+using Umbraco.Core.Services;
+using Umbraco.Web;
+#endif
+
+namespace Umbraco.Cms.Integrations.Automation.Zapier.Services
+{
+    public class UserValidationService : IUserValidationService
```
AndyButland, line 14: Nice, I like how you've wrapped this up to share to different controllers and to keep the V8/V9 logic in one place.

src/Umbraco.Cms.Integrations.Automation.Zapier/Controllers/PollingController.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            if (string.IsNullOrEmpty(username) || string.IsNullOrEmpty(password)) return null;
+
+            var isAuthorized = _userValidationService.Validate(username, password, Options.UserGroup).GetAwaiter().GetResult();
+            if (!isAuthorized) return null;
+
+            var rootNodes = _contentService.GetRootContent().Where(p => p.Published)
+                .OrderByDescending(p => p.PublishDate);
+
+            foreach (var publishedContentRoot in rootNodes)
```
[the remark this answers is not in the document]
acoumb: The endpoint is used when Zap triggers are created and a request is sent to the Umbraco instance to get a sample request that is displayed to the user in the _Test Trigger_ section. It expects an array of elements, and returning only one root node was enough to validate through localhost.

src/Umbraco.Cms.Integrations.Automation.Zapier/Services/UserValidationService.cs
acoumb: Thanks.

src/Umbraco.Cms.Integrations.Automation.Zapier/Migrations/ZapContentConfigTable.cs
acoumb: Updated.

AndyButland (commented):
Looks good.  My only remaining point is that I think the indexes you've added aren't unique ones.  They don't have to be to speed up searches, but it would help - plus it ensures data integrity here.  Even if you can't add more than one web hook per doc type, or more than one doc type with the same URL via the front-end, it's still good to enforce this in the database if that's correct for the data.

Hopefully the attributes you are using have an option to specify "unique" as well as "non-clustered", and if so, I think you could add that.

AndyButland (approved):

PR #17, opened 2022-04-19
Feature/google search console

acoumb:
Current integration uses the new Google Search Console URL Inspection API to provide a tool for accessing data outside of Search Console (more details [here](https://developers.google.com/search/blog/2022/01/url-inspection-api))

The OAuth authorization process is managed using the Google Cloud Platform app _Umbraco CMS Integrations_. For local testing against the proxy, an authorized URI record has been added to the app for `https://localhost:44364/oauth/google`.

API Reference: [https://developers.google.com/webmaster-tools/v1/urlInspection.index/inspect](https://developers.google.com/webmaster-tools/v1/urlInspection.index/inspect)

Information about the structure of the URL Inspection result: [https://developers.google.com/webmaster-tools/v1/urlInspection.index/UrlInspectionResult](https://developers.google.com/webmaster-tools/v1/urlInspection.index/UrlInspectionResult).

When authorizing with a Google account, for accurate results you must have a website property defined [here](https://search.google.com/search-console) , which must be previously validated by Google, and ownership confirmed.

The input fields are by default disabled and hold the details of the current node. For testing purposes mostly, they can be enabled using the _Edit_ action and different URLs inspected.


src/Umbraco.Cms.Integrations.SEO.GoogleSearchConsole.UrlInspectionTool/App_Plugins/UmbracoCms.Integrations/SEO/GoogleSearchConsole/URLInspectionTool/js/url-inspection-tool.controller.js
```diff
@@ -0,0 +1,115 @@
...
+    vm.loading = false;
+    vm.showResults = false;
+    vm.oauthConfiguration = {};
+    vm.inspectionResult = {};
+
+    // get available cultures for current node
+    vm.currentCultures = editorState.current.urls.map(p => p.culture);
+
+    // build default url inspection object
+    vm.siteUrl = location.href.slice(0, location.href.indexOf("umbraco") - 1);
```
AndyButland, line 13: You might have to find a way to do this by position, as it's not 100% sure that the root path is `/umbraco/` (I believe you can configure something else).

Don't forget also that the user facing website URL may be different to that used for the back-office.

src/Umbraco.Cms.Integrations.SEO.GoogleSearchConsole.UrlInspectionTool/App_Plugins/UmbracoCms.Integrations/SEO/GoogleSearchConsole/URLInspectionTool/urlInspection.html
```diff
@@ -0,0 +1,125 @@
...
+                        state="init"
+                        label="Revoke"
+                        disabled="!vm.oauthConfiguration.isConnected">
+            </umb-button>
+        </umb-box-header>
+        <umb-box-content>
+            <div>
+                <h5>About Google Search Console - URL Inspection API</h5>
+                <p>
+                    The Search Console APIs are a way to access data outside of Search Console, through external applications and products.
```
AndyButland, line 24: I'd suggest look to reduce this to just a couple of paragraphs and see if you can make it a little more editor friendly.  I.e. talking less about APIs, parameters and responses and more about relevant information about the page that is used by Google to determine the search engine ranking.

Do keep the bit about usage limits though, that's important for them to know.

src/Umbraco.Cms.Integrations.SEO.GoogleSearchConsole.UrlInspectionTool/Umbraco.Cms.Integrations.SEO.GoogleSearchConsole.URLInspectionTool.csproj
```diff
@@ -0,0 +1,71 @@
...
+	</PropertyGroup>
+
+	<PropertyGroup>
+		<PackageId>Umbraco.Cms.Integrations.SEO.GoogleSearchConsole.URLInspectionTool</PackageId>
+		<Title>Umbraco CMS Integrations: SEO - Google Search Console URL Inspection Tool</Title>
+		<Description>An extension for Umbraco CMS providing programmatic access to URL-level data for properties managed in Search Console and the indexed version of a URL.</Description>
+		<PackageIconUrl></PackageIconUrl>
+		<PackageProjectUrl>https://github.com/umbraco/Umbraco.Cms.Integrations</PackageProjectUrl>
+		<RepositoryUrl>https://github.com/umbraco/Umbraco.Cms.Integrations</RepositoryUrl>
+		<Version>1.0.0</Version>
```
AndyButland, line 13: Perhaps start with 0.1.0 so we can release and test before committing to a V1 version.

AndyButland (commented):
Looks good so far.  I added a few minor comments inline.

Doesn't look like I currently have access to test it so maybe we can do a call and you can show me and talk me through it.

Otherwise only more significant thought was making sure we can handle multiple URLs for a given node.  For example. with multi-language, we may have root paths defined so the single home page is available at `/en/` and `/it/` (or `en.mysite.com` and `it.mysite.com`.  I can see you have a "Language Code" selector but I think we may actually need that first box for "Inspection URL" being a drop-down so you can pick all the URLs defined for the node.  These will be all the ones you can see listed on the "Info" content app, under the "Links" heading.

src/Umbraco.Cms.Integrations.SEO.GoogleSearchConsole.UrlInspectionTool/Umbraco.Cms.Integrations.SEO.GoogleSearchConsole.URLInspectionTool.csproj
acoumb: Updated

src/Umbraco.Cms.Integrations.SEO.GoogleSearchConsole.UrlInspectionTool/App_Plugins/UmbracoCms.Integrations/SEO/GoogleSearchConsole/URLInspectionTool/js/url-inspection-tool.controller.js
acoumb: Updated. Included a dropdownlist to handle the different URLs, and the language code will be automatically picked by matching with the selected URL.

acoumb:
Integration can be tested on this cloud instance: [https://umbraco-sandbox.euwest01.umbraco.io](https://umbraco-sandbox.euwest01.umbraco.io)

AndyButland (approved):
Looks good now, approving... but just a couple of further comments.

Would you be able to wire up this Cloud site so that Google knows we own it, and ideally has indexed it, so we get some data back (I just get the "you don't own the site" message at the moment)?

If I had culture and host names I get these generated:

image

As you can see, really there should be only two, as the other two are effectively duplicates.  Could we de-duplicate this in the content app do you think?

acoumb:
Updated

PR #18, opened 2022-04-21
Feature/zapier integration upgrade

acoumb:
Based on the Zapier platform requirements for publishing apps, the following areas have been refactored:

1. Database table - store the content type alias and hook URL.
2. Updated migration plan.
3. Simplified the dashboard to display only info data and the subscription hooks.
4. Added endpoint to retrieve document types and display them in the dynamic field when creating the Zap trigger (once a trigger is created for a content type, it will not show up in the dynamic field, unless the user stops a Zap and the Unsubscribe action is called).
5. Subscribe/Unsubscribe are handled through the same API endpoint by sending the _subscribeHook_ flag true or false.

The integration can be tested here: [https://umbraco-sandbox.euwest01.umbraco.io/umbraco#/content](https://umbraco-sandbox.euwest01.umbraco.io/umbraco#/content)


src/Umbraco.Cms.Integrations.Automation.Zapier/Components/NewContentPublishedComponent.cs
```diff
@@ -49,22 +49,23 @@ private void ContentServiceOnPublished(IContentService sender, ContentPublishedE
...
+                        {Constants.Content.Name, node.Name},
+                        {Constants.Content.PublishDate, DateTime.UtcNow.ToString()}
+                    };
 
-                var t = Task.Run(async () => await _zapierService.TriggerAsync(zapContentConfig.WebHookUrl, content));
+                    var t = Task.Run(
+                        async () => await _zapierService.TriggerAsync(zapContentConfig.HookUrl, content));
 
-                var result = t.Result;
+                    var result = t.Result;
```
AndyButland, line 64: You could move these lines into a helper class - either a static method, or another class that you inject an instance of - so you can share the code between the V8 and V9 versions.

src/Umbraco.Cms.Integrations.Automation.Zapier/Migrations/ZapierSubscriptionHookTable.cs
```diff
@@ -34,44 +34,41 @@ public override void Migrate()
...
+        public class ZapierSubscriptionHook
         {
             [PrimaryKeyColumn(AutoIncrement = true, IdentitySeed = 1)]
             [Column("Id")]
             public int Id { get; set; }
 
-            [Column("ContentTypeName")]
-            [Index(IndexTypes.UniqueNonClustered, Name = "IX_ZapContentConfig_ContentTypeName")]
-            public string ContentTypeName { get; set; }
+            [Column("ContentTypeAlias")]
```
AndyButland, line 65: Just thinking about how we'd extend this for other triggers, such as form submissions.  Would we create a new table?  Or reuse this table?

If the latter - and I think that makes sense - we would need to have this field more generically named, so it could store either a content type alias, or a form Guid, as appropriate.  So perhaps `EntityIdentifier`?

src/Umbraco.Cms.Integrations.Automation.Zapier/Controllers/ContentController.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            if (string.IsNullOrEmpty(username) || string.IsNullOrEmpty(password)) return Enumerable.Empty<ContentTypeDto>();
+
+            var isAuthorized = _userValidationService.Validate(username, password, Options.UserGroup).GetAwaiter().GetResult();
+            if (!isAuthorized) return Enumerable.Empty<ContentTypeDto>();
+
+            var contentTypes = _contentTypeService.GetAll();
+
+            return contentTypes
+                .Where(p => !_zapierSubscriptionHookService.TryGetByAlias(p.Alias, out var contentConfig))
```
AndyButland, line 82: Do we need to restrict here?  I'm just thinking it might be OK for someone (or two different editors) to create two triggers on the same doc type.  E.g. one wants to send a tweet and the other one an email.

AndyButland (approved):
Looks good to me.  A few comments inline.

src/Umbraco.Cms.Integrations.Automation.Zapier/App_Plugins/UmbracoCms.Integrations/Automation/Zapier/dashboard.html
```diff
@@ -22,50 +23,33 @@
                     You can initiate your automation when a content item of a particular document type is published in Umbraco.
```
AndyButland, line 23: It would be nice if we could update this to say "... or a form is submitted on your website" if the forms integration is also installed.  I.e. if `vm.formsExtensionInstalled` is `true`.

src/Umbraco.Cms.Integrations.Automation.Zapier/App_Plugins/UmbracoCms.Integrations/Automation/Zapier/dashboard.html
```diff
@@ -22,50 +23,33 @@
...
                 </p>
                 <p>
-                    Using the filters below, map content items with Zap triggered webhooks.  This will:
+                    The integration uses Zapier subscription hook triggers, allowing Zapier to set up and remove hook subscriptions through Subscribe/Unsubscribe API requests.
```
AndyButland, line 26:
```suggestion
                    The integration uses Zapier subscription hook triggers, allowing Zapier to set up and remove hook subscriptions when Zaps are created or removed on the platform.
```

Just a little less technical.

src/Umbraco.Cms.Integrations.Automation.Zapier/Helpers/FormsHelper.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Linq;
+using System.Reflection;
+
+namespace Umbraco.Cms.Integrations.Automation.Zapier.Helpers
+{
+    public class FormsHelper
+    {
+        public static bool IsFormsExtensionInstalled => Type.GetType("Umbraco.Forms.Core.Services.IFormService, Umbraco.Forms.Core") != null;
+
+        public static IEnumerable<MethodInfo> GetMethodsForType(Type type)
```
AndyButland, line 12: I'd suggest moving this method and the one below into a new class, e.g. `ReflectionHelper`, just as they don't really have anything to do with Forms and are more utility methods.

src/Umbraco.Cms.Integrations.Automation.Zapier/Components/NewContentPublishedNotification.cs
```diff
@@ -30,24 +29,27 @@ public NewContentPublishedNotification(ZapConfigService zapConfigService, Zapier
...
-
-                var result = t.Result;
-
-                if (!string.IsNullOrEmpty(result))
-                    _logger.LogError(result);
+                    var content = new Dictionary<string, string>
+                    {
+                        {Constants.Content.Id, node.Id.ToString() },
+                        {Constants.Content.Name, node.Name },
+                        {Constants.Content.PublishDate, DateTime.UtcNow.ToString("s") }
```
AndyButland, line 42: This still looks to be sending only the ID, name and publish date of the content, but I was expecting now we would be sending all the field values of the content too.

src/Umbraco.Cms.Integrations.Automation.Zapier/Controllers/ContentController.cs
```diff
@@ -0,0 +1,91 @@
...
+#else
+using System.Configuration;
+
+using Umbraco.Web.WebApi;
+using Umbraco.Core.Services;
+#endif
+
+namespace Umbraco.Cms.Integrations.Automation.Zapier.Controllers
+{
+    public class ContentController : UmbracoApiController
```
AndyButland, line 24: Just to help understand the flow, think we could do with some XML header comments on this and the other various controllers.  I.e. for this one, explaining that it exposes and end-point to Zapier so it can create a dynamic drop-down list for the content types to pick from.

src/Umbraco.Cms.Integrations.Automation.Zapier/Controllers/PollingController.cs
```diff
@@ -39,12 +50,16 @@ public PollingController(IContentService contentService, IUserValidationService
...
+
             _contentService = contentService;
 
+            _zapierFormService = zapierFormService;
+
             _userValidationService = userValidationService;
         }
 
-        public List<Dictionary<string, string>> GetContent()
+        public IEnumerable<PublishedContentDto> GetSampleContent()
```
AndyButland, line 62: I'm a bit confused about this controller.  `GetSampleContent` and `GetSampleForm` don't seem to be consistent.  By that I mean the former is returning a partial definition of some published content, but the latter all the forms.  Seems to be if anything it should be exposing form submissions, not the form definitions (i.e. what would be the content of the zap information).

And `GetContentByAlias` is a concern as it looks like it's getting all the published content on the site for a given content type, which could be a lot and I don't think we'd want to respond to that on a poll.

src/Umbraco.Cms.Integrations.Automation.Zapier/Services/ZapierFormService.cs
```diff
@@ -0,0 +1,92 @@
...
+#if NETFRAMEWORK
+using System.Web.Mvc;
+#endif
+
+using Umbraco.Cms.Integrations.Automation.Zapier.Helpers;
+using Umbraco.Cms.Integrations.Automation.Zapier.Models.Dtos;
+
+namespace Umbraco.Cms.Integrations.Automation.Zapier.Services
+{
+    public class ZapierFormService
```
AndyButland, line 14: You could move this (and the `FormController`) to the forms integration, and then can avoid using reflection.  Up to you now you've done it, but given we need it for the event handling we could simplify retrieving the forms as well.

The only thing you'd need to keep here in this integration is `FormsHelper.IsFormsExtensionInstalled()`.

AndyButland (changes requested):
Some more comments inline on the latest.  Looks good overall and can see it's coming together, but to summarise the main comments think we still need to consider:

- What's best to do in the "polling" context/
- Make sure we are sending data about the fields of the content in the notification
- Consider moving the forms code out into the forms integration.

src/Umbraco.Cms.Integrations.Automation.Zapier/App_Plugins/UmbracoCms.Integrations/Automation/Zapier/dashboard.html
acoumb: Updated.

src/Umbraco.Cms.Integrations.Automation.Zapier/Helpers/FormsHelper.cs
acoumb: Updated.

src/Umbraco.Cms.Integrations.Automation.Zapier/Components/NewContentPublishedNotification.cs
acoumb: Fixed.

src/Umbraco.Cms.Integrations.Automation.Zapier/Controllers/ContentController.cs
acoumb: Updated.

src/Umbraco.Cms.Integrations.Automation.Zapier/Controllers/PollingController.cs
acoumb: Updated.

src/Umbraco.Cms.Integrations.Automation.Zapier/Services/ZapierFormService.cs
acoumb: Updated.

acoumb:
In addition, the _PollingController_ keeps the initial _GetSampleContent_ that is used by version 1.0.0 of the Zapier Umbraco App.

src/Umbraco.Cms.Integrations.Automation.Zapier/Components/NewContentPublishedComponent.cs
```diff
@@ -51,19 +49,11 @@ private void ContentServiceOnPublished(IContentService sender, ContentPublishedE
 
             foreach (var node in e.PublishedEntities)
             {
+                var x = node.ToContentDictionary();
```
AndyButland, line 52: Debug code?  To remove if so.

src/Umbraco.Cms.Integrations.Automation.Zapier/Controllers/AuthController.cs
```diff
@@ -19,6 +19,9 @@
 
 namespace Umbraco.Cms.Integrations.Automation.Zapier.Controllers
 {
+    /// <summary>
+    /// When a Zapier user creates triggers using the Umbraco app from the Zapier App Directory, he needs to provide valid credentials for a backoffice account.
```
AndyButland, line 23: "he needs" --> "they need"

src/Umbraco.Cms.Integrations.Automation.Zapier/Controllers/ContentController.cs
```diff
@@ -21,6 +20,9 @@
 
 namespace Umbraco.Cms.Integrations.Automation.Zapier.Controllers
 {
+    /// <summary>
+    /// When a Zapier user creates a new "New Content Published" trigger, the API is used to provide him with the list of content types for handling "Published" event.
```
AndyButland, line 24: remove "him with"

src/Umbraco.Cms.Integrations.Automation.Zapier/Controllers/PollingController.cs
```diff
@@ -12,20 +13,21 @@
...
 using Umbraco.Web.WebApi;
 using Umbraco.Core.Services;
-using Umbraco.Core.Models.PublishedContent;
-
 #endif
 
 namespace Umbraco.Cms.Integrations.Automation.Zapier.Controllers
 {
+    /// <summary>
+    /// When a Zapier user creates a "New Content Published" triggered, he is authenticated, then selects a content type, the API provides an output json with the
```
AndyButland, line 27: "he is" --> "they are", "then selects" --> "then select"

src/Umbraco.Cms.Integrations.Automation.Zapier/Controllers/PollingController.cs
```diff
@@ -97,43 +96,25 @@ public IEnumerable<PublishedContentDto> GetSampleContent()
...
+        public List<Dictionary<string, string>> GetContentByType(string alias)
         {
             var list = new List<Dictionary<string, string>>();
 
             using (var cref = _umbracoContextFactory.EnsureUmbracoContext())
             {
                 var cache = cref.UmbracoContext.Content;
                 
-                var nodes = cache.GetByXPath($"//{contentTypeAlias}")
+                var node = cache.GetByXPath($"//{alias}")
```
AndyButland, line 107: This doesn't look quite right to me still.  Would we not be better here using the `IContentTypeService` to retrieve the content type (not the content of a provided type, which may not exist)?

When we have that we can format a response (using the current date for the published time, and empty strings for the property values).

AndyButland (commented):
I've reviewed the latest.  Just one comment really other than some small inline things around how we return the sample response for a given content type to Zapier.

src/Umbraco.Cms.Integrations.Automation.Zapier/Components/NewContentPublishedComponent.cs
acoumb: Fixed.

src/Umbraco.Cms.Integrations.Automation.Zapier/Controllers/ContentController.cs
acoumb: Updated.

src/Umbraco.Cms.Integrations.Automation.Zapier/Controllers/PollingController.cs
acoumb: Updated.

src/Umbraco.Cms.Integrations.Automation.Zapier/Controllers/PollingController.cs
acoumb: Updated.

src/Umbraco.Cms.Integrations.Automation.Zapier/Controllers/PollingController.cs
```diff
@@ -98,23 +59,13 @@ public IEnumerable<PublishedContentDto> GetSampleContent()
...
-                var cache = cref.UmbracoContext.Content;
-                
-                var node = cache.GetByXPath($"//{alias}")
-                    .Where(p => p.IsPublished())
-                    .OrderByDescending(p => p.UpdateDate)
-                    .FirstOrDefault();
+            if (!IsUserValid()) return null;
 
-                if (node == null) return list;
+            var contentType = _contentTypeService.Get(alias);
```
AndyButland, line 64: Should have a null check here I think, in case the content type isn't found.  It would be unexpected, but still should consider here - and I guess return an empty dictionary.

AndyButland (approved):
Looks good and ready for merging now.  With just one small update suggested first.

PR #19, opened 2022-05-17
Feature/google compliance

acoumb:
Replaced Connect button with assets received from Google Support team, and appearing now as _Google Signin_


AndyButland (approved):
Looks good (and on brand!) to me.

PR #21, opened 2022-05-24
Migration update

acoumb:
Included changes:

- merge content/form subscriptions into a single table.
- reference entity by content type and form id
- updated dashboard


src/Umbraco.Cms.Integrations.Automation.Zapier/Migrations/ZapierMigration.cs
```diff
@@ -80,31 +60,18 @@ public class ZapierSubscriptionHookTable
             [Column("Id")]
             public int Id { get; set; }
 
-            [Column("ContentTypeAlias")]
-            [Index(IndexTypes.UniqueNonClustered, Name = "IX_ZapierSubscriptionHook_ContentTypeAlias")]
-            public string ContentTypeAlias { get; set; }
+            /// <summary>
+            /// Column stores two types of references:
+            /// 1. content type for content triggers
```
AndyButland, line 65:
```suggestion
            /// 1. content type alias for content triggers
```

src/Umbraco.Cms.Integrations.Automation.Zapier/Models/Dtos/SubscriptionDto.cs
```diff
@@ -1,16 +1,24 @@
...
         public string HookUrl { get; set; }
 
-        [JsonProperty("contentType")]
-        public string ContentType { get; set; }
-
         [JsonProperty("subscribeHook")]
         public bool SubscribeHook { get; set; }
+
+        [JsonProperty("isFormSubscription")]
+        public bool IsFormSubscription => Guid.TryParse(EntityId, out _);
```
AndyButland, line 22: Thinking here we can make the package more extensible by removing this and adding in a "Type" field.  It should take an integer, and have 1 for content publishing and 2 for form submissions.  We can then more easily add more in future.

There should be a unique index across the EntityId and Type fields.

When you look up an entity, you should look it up by type and entity Id (just in case in future we had two types with the same Id).

And perhaps you can expose this on the dashboard, so you only need a single table for display.

acoumb: Updated

src/Umbraco.Cms.Integrations.Automation.Zapier/Migrations/ZapierMigration.cs
```diff
@@ -62,13 +62,17 @@ public class ZapierSubscriptionHookTable
...
-            /// 1. content type for content triggers
+            /// 1. content type alias for content triggers
             /// 2. form ID for form triggers
             /// </summary>
             [Column("EntityId")]
             [Index(IndexTypes.UniqueNonClustered, Name = "IX_ZapierSubscriptionHook_EntityId")]
             public string EntityId { get; set; }
 
+            [Column("Type")]
+            [Index(IndexTypes.UniqueNonClustered, Name = "IX_ZapierSubscriptionHook_Type", ForColumns = "EntityId,Type")]
```
AndyButland, line 73: Now you've added this index, you should remove the one on `EntityId` (as that, theoretically, is no longer unique by itself).

src/Umbraco.Cms.Integrations.Automation.Zapier/Models/Dtos/EntityDto.cs
```diff
@@ -0,0 +1,12 @@
+﻿
+namespace Umbraco.Cms.Integrations.Automation.Zapier.Models.Dtos
+{
+    public class EntityDto
+    {
+        public enum Type
+        {
+            Content = 1,
```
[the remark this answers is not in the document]
acoumb: updated.

PR #23, opened 2022-05-26
V9/bugfix/partial view missing

acoumb:
Updates included with version 1.1.0 of the package:
- rendering views for V8 and V9 with html mockup
- updated helper method allowing the user to use it's own view


AndyButland (approved):
Looks fine to me.  I'd just suggest going to `1.0.1` as the version update, as really this is a patch (it's fixing an issue rather than adding a feature, and there are no breaking changes).

PR #24, opened 2022-06-02
Validate access using also an API key.

acoumb:
Additional feature to control platform access through an API key.

Use cases:

1. User tries to authenticate when a Zap trigger is created: if an API key is provided in the POST request, the access is validated by comparing the request value with an API key value saved in the configuration file of the website; otherwise, the access validation takes into consideration the user credentials.
2. Polling and Subscribe actions are called: the same validation process as above runs.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Automation.Zapier/Services/UserValidationService.cs
+++ b/src/Umbraco.Cms.Integrations.Automation.Zapier/Services/UserValidationService.cs
@@ -1,10 +1,16 @@
 ﻿using System.Linq;
 using System.Threading.Tasks;
 
+using Umbraco.Cms.Integrations.Automation.Zapier.Configuration;
+
 #if NETCOREAPP
+using Microsoft.Extensions.Options;
+
 using Umbraco.Cms.Core.Security;
 using Umbraco.Cms.Core.Services;
 #else
+using System.Configuration;
+
 using Umbraco.Core.Services;
 #endif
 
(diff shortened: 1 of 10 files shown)


src/Umbraco.Cms.Integrations.Automation.Zapier/Services/UserValidationService.cs
```diff
@@ -40,11 +61,11 @@ public async Task<bool> Validate(string username, string password, string userGr
 
             if (!isUserValid) return false;
 
-            if (!string.IsNullOrEmpty(userGroup))
+            if (!string.IsNullOrEmpty(_zapierSettings.UserGroup))
             {
                 var user = _userService.GetByUsername(username);
 
-                return user != null && user.Groups.Any(p => p.Name == userGroup);
+                return user != null && user.Groups.Any(p => p.Name == _zapierSettings.UserGroup);
```
AndyButland, line 68: Just spotted this, but we should probably be using the user group alias rather than the name here, as that's less likely to change and then break an integration.

AndyButland (approved):
LGTM.  Just one comment (which wasn't from this update, but I only spotted it this time).

src/Umbraco.Cms.Integrations.Automation.Zapier/Services/UserValidationService.cs
acoumb: Updated

Issue #31, opened 2022-07-28
Add Shopify product id


richard-arkom:
Would be really useful if someone could add the Shopify product Id to the ShopifyProductPickerValueConverter / ProductViewModel so we can access it on our pages.

Don't think it would take very long.

Cheers

richard-arkom:
Hi,

Thank you for adding in the Product Id field, it will be so useful.

Quick question, how long is it before the version available in the NuGet Package Manager will include the extra field, or is there another way I can install the new version.

Cheers

acoumb:
Hi Richard,

I have released version [1.0.2](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Commerce.Shopify/1.0.2) of the package that includes the _Product Id_ and an update to the asynchronous call to Shopify API in value converter.

I will now close the issue, thank you for your feedback!

PR #32, opened 2022-08-03
Issue #31 - return product id

acoumb:
This PR handles the issue reported [here](https://github.com/umbraco/Umbraco.Cms.Integrations/issues/31), returning the Shopify product id with the view model.


AndyButland (approved):
Looks good to me.  Only thing I thought was that we don't really document this model, and perhaps we should.

I think it would suffice if you added a link from the phrase "strongly typed model" under [this section](https://github.com/umbraco/Umbraco.Cms.Integrations/blob/main/src/Umbraco.Cms.Integrations.Commerce.Shopify/readme.md#front-end-rendering) of the read me, to [this code file](https://github.com/umbraco/Umbraco.Cms.Integrations/blob/main/src/Umbraco.Cms.Integrations.Commerce.Shopify/Models/ViewModels/ProductViewModel.cs).

AndyButland:
Would just note the build is failing, so that may need looking into.  Not related to this PR though.

Issue #30, opened 2022-07-19
Hubspot authentication blocks editor


FransdeJong:
If you use the form picker for Hubspot with OAuth you need rights to the datatype to login to hubspot and pick forms.

Editors without settings rights can't add form without developer help. This seems a bit unfriendly.
Would it be possible to, instead of throwing only a warning, to enable editors to login from the property editor?

acoumb:
Hi Frans, thank you for your feedback. I have released version [1.1.2](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Crm.Hubspot/1.1.2) of the package covering your use case.

PR #34, opened 2022-08-09
acoumb

acoumb:
This PR handles the issue reported [here](https://github.com/umbraco/Umbraco.Cms.Integrations/issues/30) and allows editors to control OAuth access from the property editor also, not just from the data type, using a custom implemented directive.


src/Umbraco.Cms.Integrations.Crm.Hubspot/Umbraco.Cms.Integrations.Crm.Hubspot.csproj
```diff
@@ -43,10 +43,18 @@
 		</Content>
 	</ItemGroup>
 
+	<ItemGroup>
```
AndyButland, line 46: Don't think this needed to be added.

AndyButland (approved):
Looks good to me.

PR #36, opened 2022-08-11
Packages migration to v10 & Umbraco 10 Test Site

acoumb:


acoumb:
Removed .net 6 as targeting framework as tests confirmed the packages work in their current state.

AndyButland (approved):

PR #44, opened 2022-09-12
Use cases updates

acoumb:
Current PR fixes a couple of use cases that create inconsistencies in the normal workflow of the Dynamics extension:

1. Forms array contains duplicates as the push action is managed twice.
2. If a form is not live in one Dynamics instance, an exception was thrown with not too much information for the user. Now a warning message will be prompted to the editor, notifying him that he should revise the form status.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Crm.Dynamics/App_Plugins/UmbracoCms.Integrations/Crm/Dynamics/js/formpicker.controller.js
+++ b/src/Umbraco.Cms.Integrations.Crm.Dynamics/App_Plugins/UmbracoCms.Integrations/Crm/Dynamics/js/formpicker.controller.js
@@ -79,8 +79,8 @@
                     });
                 });
             }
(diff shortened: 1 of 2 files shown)


src/Umbraco.Cms.Integrations.Crm.Dynamics/App_Plugins/UmbracoCms.Integrations/Crm/Dynamics/js/formpicker.controller.js
```diff
@@ -77,6 +80,7 @@
                 });
             }
         });
+        vm.loading = false;
```
AndyButland, line 83: Looks like this should be within the callback function, not outside it, otherwise it will execute immediately.

AndyButland (commented):
Just one comment that looks like it needs correcting, otherwise looks good.

Issue #50, opened 2022-09-30
Issues connecting to UK D365 Marketing API in Umbraco 10


martinjgriffiths:
I am having problems connecting to our production instance of D365M.

Host URL: https://[hostname].api.crm11.dynamics.com/

ApiPath: api/data/v9.2/

I can connect via oAuth/AD but I am getting the following error after submitting my creds...

An error occurred
Invalid URI: The format of the URI could not be determined.

Exception Details
System.UriFormatException, System.Private.Uri, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a: Invalid URI: The format of the URI could not be determined.

![image](https://user-images.githubusercontent.com/883638/193307932-46ac1c27-59aa-4028-9607-d91d86f80675.png)

acoumb:
Hi Martin,
I will review this and see what is the issue.
Thank you

acoumb:
In the meantime, could you confirm that the Host URL and API Path are the right ones?
This section from the [blog post](https://umbraco.com/blog/integrating-umbraco-cms-with-microsoft-dynamics-365-marketing-forms/):
 _To access the API, you will need to retrieve the HTTP REST API details of your Dynamics 365 instance following these steps:_

acoumb:
Hi @martinjgriffiths . Did you manage to test the new version of the package? 
I am closing this for now, but feel free to re-open if you have other feedback.

PR #51, opened 2022-10-04
Validate users outside organisation; HTTP revoke action update.

acoumb:
Updates included in this PR cover the issue described [here](https://github.com/umbraco/Umbraco.Cms.Integrations/issues/50):

- exception thrown when user is authenticated. An use case when this can be replicated is when an account outside the organization tries to authenticate, then an empty string is sent to the Microsoft API to retrieve user details, at which point the exception is thrown. Added validation to check and display an error message to the user if this occurs.
- _Revoke_ action throws a _400 Bad Request_ error on the POST action, so I changed it to a DELETE Http method.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Crm.Dynamics/Models/Dtos/IdentityDto.cs
+++ b/src/Umbraco.Cms.Integrations.Crm.Dynamics/Models/Dtos/IdentityDto.cs
@@ -4,12 +4,14 @@ namespace Umbraco.Cms.Integrations.Crm.Dynamics.Models.Dtos
 {
     public class IdentityDto
     {
-        public bool IsAuthorized { get; set; } = true;
+        public bool IsAuthorized { get; set; }
 
         [JsonProperty("systemuserid")]
         public string UserId { get; set; }
 
         [JsonProperty("fullname")]
         public string FullName { get; set; }
+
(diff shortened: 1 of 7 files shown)


src/Umbraco.Cms.Integrations.Crm.Dynamics/Models/Dtos/IdentityDto.cs
```diff
@@ -4,12 +4,14 @@ namespace Umbraco.Cms.Integrations.Crm.Dynamics.Models.Dtos
 {
     public class IdentityDto
     {
-        public bool IsAuthorized { get; set; } = true;
+        public bool IsAuthorized { get; set; }
```
AndyButland, line 7: Are you missing a `[JsonProperty("isAuthorized")]` here?

AndyButland (approved):
Looks fine to me if it sorts out the issues you found.

src/Umbraco.Cms.Integrations.Crm.Dynamics/Models/Dtos/IdentityDto.cs
acoumb: This flag is only used server-side, for the client I am using the _IsAuthorized_ serializable property of _OAuthConfigurationDto_ .

PR #52, opened 2022-10-12
V10/feature/activecampaign integration

acoumb:
Integration provides a form picker and rendering component for ActiveCampaign.
Targeted version is Umbraco 10 and .NET6.


src/Umbraco.Cms.Integrations.Crm.ActiveCampaign/Controllers/FormsController.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            var response = await client.SendAsync(new HttpRequestMessage { Method = HttpMethod.Get });
+
+            var content = await response.Content.ReadAsStringAsync();
+
+            if (!response.IsSuccessStatusCode)
+                return new ResponseDto<IEnumerable<FormDto>>
+                {
+                    Data = Enumerable.Empty<FormDto>(),
+                    Message = String.IsNullOrEmpty(content)
```
AndyButland, line 44:
```suggestion
                    Message = string.IsNullOrEmpty(content)
```

src/Umbraco.Cms.Integrations.Crm.ActiveCampaign/Controllers/FormsController.cs
```diff
@@ -0,0 +1,92 @@
...
+                            ? Constants.Resources.AuthorizationFailed : Constants.Resources.ApiAccessFailed
+                        : JsonNode.Parse(content)["message"].ToString()
+                };
+
+            return new ResponseDto<IEnumerable<FormDto>>
+            {
+                Data = string.IsNullOrEmpty(content)
+                    ? Enumerable.Empty<FormDto>()
+                    : JsonNode.Parse(content)["forms"].Deserialize<IEnumerable<FormDto>>(),
+                Message = String.Empty
```
AndyButland, line 55:
```suggestion
                Message = string.Empty
```

src/Umbraco.Cms.Integrations.Crm.ActiveCampaign/Controllers/FormsController.cs
```diff
@@ -0,0 +1,92 @@
...
+                    Method = HttpMethod.Get,
+                    RequestUri = new Uri(client.BaseAddress + "/" + id)
+                });
+
+            var content = await response.Content.ReadAsStringAsync();
+
+            if (!response.IsSuccessStatusCode)
+                return new ResponseDto<FormDto>
+                {
+                    Message = String.IsNullOrEmpty(content)
```
AndyButland, line 76:
```suggestion
                    Message = string.IsNullOrEmpty(content)
```

src/Umbraco.Cms.Integrations.Crm.ActiveCampaign/Controllers/FormsController.cs
```diff
@@ -0,0 +1,92 @@
...
+                        ? response.StatusCode == System.Net.HttpStatusCode.Forbidden
+                            ? Constants.Resources.AuthorizationFailed : Constants.Resources.ApiAccessFailed
+                        : JsonNode.Parse(content)["message"].ToString()
+                };
+
+            return new ResponseDto<FormDto>
+            {
+                Data = string.IsNullOrEmpty(content)
+                ? new FormDto()
+                : JsonNode.Parse(content)["form"].Deserialize<FormDto>(),
```
AndyButland, line 86:
```suggestion
                    ? new FormDto()
                    : JsonNode.Parse(content)["form"].Deserialize<FormDto>(),
```

src/Umbraco.Cms.Integrations.Crm.ActiveCampaign/Models/Dtos/ApiAccessDto.cs
```diff
@@ -0,0 +1,25 @@
+﻿using Newtonsoft.Json;
+
+namespace Umbraco.Cms.Integrations.Crm.ActiveCampaign.Models.Dtos
+{
+    public class ApiAccessDto
+    {
+        private string _baseUrl { get; set; }
+
+        private string _apiKey { get; set; }
```
AndyButland, line 9:
```suggestion
        private readonly string _baseUrl;

        private readonly string _apiKey;
```

I think we can just use private fields here rather than properties, and they can be readonly as they are only set in the constructor.

src/Umbraco.Cms.Integrations.Crm.ActiveCampaign/Models/Dtos/FormDto.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Newtonsoft.Json;
+using System.Text.Json.Serialization;
+
+namespace Umbraco.Cms.Integrations.Crm.ActiveCampaign.Models.Dtos
+{
+    public class FormDto
+    {
+        [JsonProperty("id")]
```
AndyButland, line 8: Do we need both these attributes here?  Seems we should be using `Newtonsoft.Json` or `System.Text.Json` throughout, but not both.

src/Umbraco.Cms.Integrations.Crm.ActiveCampaign/Models/Dtos/ResponseDto.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Newtonsoft.Json;
+
+using System.Text.Json.Serialization;
+
+namespace Umbraco.Cms.Integrations.Crm.ActiveCampaign.Models.Dtos
+{
+    public class ResponseDto<T>
+        where T : class
+    {
+        [JsonProperty("data")]
```
AndyButland, line 10: Also here the point about using `Newtonsoft.Json` or `System.Text.Json` throughout, but not both.

src/Umbraco.Cms.Integrations.Crm.ActiveCampaign/Umbraco.Cms.Integrations.Crm.ActiveCampaign.csproj
```diff
@@ -0,0 +1,28 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<TargetFramework>net6.0</TargetFramework>
+		<ImplicitUsings>enable</ImplicitUsings>
+		<Nullable>enable</Nullable>
```
AndyButland, line 6: You could add `<WarningsAsErrors>Nullable</WarningsAsErrors>` here - which will ensure you are correctly handling nullable types.

src/Umbraco.Cms.Integrations.Crm.ActiveCampaign/Umbraco.Cms.Integrations.Crm.ActiveCampaign.csproj
```diff
@@ -0,0 +1,28 @@
...
+		<PackageId>Umbraco.Cms.Integrations.Crm.ActiveCampaign</PackageId>
+		<Title>Umbraco CMS Integrations: CRM - ActiveCampaign</Title>
+		<Description>An extension for Umbraco CMS providing a form picker and rendering component for ActiveCampaign.</Description>
+		<PackageIconUrl></PackageIconUrl>
+		<PackageProjectUrl>https://github.com/umbraco/Umbraco.Cms.Integrations</PackageProjectUrl>
+		<RepositoryUrl>https://github.com/umbraco/Umbraco.Cms.Integrations</RepositoryUrl>
+		<Version>1.0.0</Version>
+		<Authors>Umbraco HQ</Authors>
+		<Company>Umbraco</Company>
+		<ApplicationManifest>App_Plugins\UmbracoCms.Integrations\Crm\ActiveCampaign\package.manifest</ApplicationManifest>
```
AndyButland, line 19: Do we need this here?  I've not seen it before so just flagging in case you are mixing two concepts here, Umbraco's package manifests and a Windows executables one.

AndyButland (approved):
Looks great.  A few comments inline, but all minor.

I'm approving, but first can we arrange a call so you can demo it in action to me too please?

acoumb:
Additional updates included:

- notification used in the form picker editor instead of the directive.
- save form id instead of an object and pull form details using the API.
- small UI updates.

src/Umbraco.Cms.Integrations.Crm.ActiveCampaign/App_Plugins/UmbracoCms.Integrations/Crm/ActiveCampaign/js/configuration.controller.js
```diff
@@ -0,0 +1,16 @@
+﻿function configurationController($scope, notificationsService, umbracoCmsIntegrationsCrmActiveCampaignResource) {
+    var vm = this;
+
+    umbracoCmsIntegrationsCrmActiveCampaignResource.checkApiAccess().then(function (response) {
+
+        if (response.isApiConfigurationValid)
+            vm.account = response.account;
+
+        vm.status = response.isApiConfigurationValid 
+            ? "Connected. Account name: "
```
AndyButland, line 10: This doesn't look quite right - aren't you missing concatenating the account name?

src/Umbraco.Cms.Integrations.Crm.ActiveCampaign/App_Plugins/UmbracoCms.Integrations/Crm/ActiveCampaign/js/formpicker.controller.js
```diff
@@ -0,0 +1,93 @@
...
+            }
+        };
+
+        editorService.open(options);
+    };
+
+    function getFormDetails(id) {
+        vm.loading = true;
+        umbracoCmsIntegrationsCrmActiveCampaignResource.getForm(id).then(function (response) {
+            if (response.message !== null && response.message.length > 0)
```
AndyButland, line 63:
```suggestion
            if (response.message && response.message.length > 0)
```

This way you deal with undefined as well as null.

src/Umbraco.Cms.Integrations.Crm.ActiveCampaign/Models/Dtos/ResponseDto.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Text.Json.Serialization;
+
+namespace Umbraco.Cms.Integrations.Crm.ActiveCampaign.Models.Dtos
+{
+    public class ResponseDto
```
AndyButland, line 5: Am thinking this seems a little odd having both the single and collection of forms response mapped to a single object.  Would it be better to have `FormResponseDto` and `FormCollectionResponseDto`, and map to the appropriate one

`FormCollectionResponseDto` would have a single property which would be a collection of `FormDto`.

`FormResponseDto` would have a single `FormDto`.

Both could inherit from a base type that defines `Message`.

AndyButland (commented):
Looks good, but I noticed a couple of further things that could perhaps be improved.

src/Umbraco.Cms.Integrations.Crm.ActiveCampaign/App_Plugins/UmbracoCms.Integrations/Crm/ActiveCampaign/js/configuration.controller.js
AndyButland: Understood - thanks.

Issue #47, opened 2022-09-23
Zapier integration unable to load Content Type data in v10


lukehook:
Umbraco v10.2.0
Umbraco.Cms.Integrations.Automation.Zapier v1.0.3

When adding Zapier integration to a fresh install hosted on Azure App Service the Zapier trigger fails to load the 'Content Type' data.

Steps to reproduce

- Create a fresh Umbraco v10.2.0 install
- Install Umbraco.Cms.Integrations.Automation.Zapier v1.0.3 nuget package
- Publish/Deploy to an azure web app (so that we get a base URL)
- Add Zapier integration for Umbraco (Either login credentials or API Key)
- Attempt to set up New Content Published trigger
- Should see an error when choosing content type "Unable to load choices
We're having trouble loading 'Content Type' data. maximum redirect reached at: {baseURL}/Account/AccessDenied?ReturnUrl=%2F"

It appears that the Zapier app is failing to pass authentication possibly.

Running a 'test' on the Umbraco connection passes successfully so not sure if this is something specific to the call for Content Types

acoumb:
Hi Luke, 
Thank you for bringing this up to our attention, I will try to replicate the issue and I will get back with more details.

lukehook:
Thank you for the prompt respone @acoumb, If I can provide any more info let me know (likewise if you discover that it works and I'm doing something stupid please do let me know 😄)

acoumb:
Hello again @lukehook ,
I replicated your use case using an Azure App Service, and I was able to query the CMS data from my Zap. 
The URL of my Umbraco 10.2.0 site is https://zapierazureapp20220927110307.azurewebsites.net and I was using this API key: _Y8uB84PMv1x%C%rw0u&i_. This video demonstrates my test case https://drive.google.com/file/d/1IaqrfDmXbBrfNwEyBRmnTDVYuadtbcSY/view?usp=sharing.
Could you test a Zap against my test site and see if it works?
Then, would you be able to share with me the URL of your site, with a test API key so I can have a look? 
Thank you

acoumb:
@lukehook that sounds great. Regarding your request, I will add it to the backlog and include it soon with a new version of the package.

acoumb:
@lukehook could you tell me what published property you are using on your document type?

lukehook:
Sorry @acoumb I completely missed this reply in my emails. Is this for the Image mentioned above do you mean? If so I was using `Umbraco.MediaPicker3`

PR #57, opened 2022-11-02
Get URL for media content types

acoumb:
This PR handles an issue reported [here](https://github.com/umbraco/Umbraco.Cms.Integrations/issues/47) for media pickers, when the string representation of the object was returned on polling.
With this update I've took into consideration the two media pickers with aliases: _Umbraco.MediaPicker3_ and _Umbraco.MediaPicker_(legacy).

Because the package requires a minimum version of 8.5.4, I've moved the constants locally, as _MediaPicker3_ was introduced with 8.14.0 and could not access it directly from _Umbraco.Core.Constants.PropertyEditors.Aliases.MediaPicker_.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Automation.Zapier/Constants.cs
+++ b/src/Umbraco.Cms.Integrations.Automation.Zapier/Constants.cs
@@ -42,12 +42,5 @@ namespace Umbraco.Cms.Integrations.Automation.Zapier
 
             public const int Form = 2;
         }
-
-        public static class MediaAliases
-        {
-            public const string UmbracoMediaPicker = "Umbraco.MediaPicker";
-
-            public const string UmbracoMediaPicker3 = "Umbraco.MediaPicker3";
-        }
     }
 }
--- a/src/Umbraco.Cms.Integrations.Automation.Zapier/Constants.cs
+++ b/src/Umbraco.Cms.Integrations.Automation.Zapier/Constants.cs
@@ -42,5 +42,12 @@ namespace Umbraco.Cms.Integrations.Automation.Zapier
 
             public const int Form = 2;
         }
+
+        public static class MediaAliases
+        {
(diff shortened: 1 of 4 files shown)


src/Umbraco.Cms.Integrations.Automation.Zapier/Umbraco.Cms.Integrations.Automation.Zapier.csproj
```diff
@@ -15,6 +15,7 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net472'">
     <PackageReference Include="UmbracoCms.Web" Version="8.5.4" />
+	  <PackageReference Include="UmbracoCms.Core" Version="8.14.1" />
```
AndyButland, line 18: I don't think it makes sense to have mixed versions here.  If we need 8.14,1 we should update the version of the existing `UmbracoCms.Web` reference,, bump the version of the package to a new minor, and make sure we set the dependency correctly when uploading the zip file to our.umbraco.com.

AndyButland, line 18: I.e.:

```suggestion
    <PackageReference Include="UmbracoCms.Web" Version="8.14.1" />
```

src/Umbraco.Cms.Integrations.Automation.Zapier/Constants.cs
```diff
@@ -42,5 +42,12 @@ public static class EntityType
 
             public const int Form = 2;
         }
+
+        public static class MediaAliases
```
AndyButland, line 46: If you need to bump the version anyway (see separate comment), you can likely reference these constants from the core CMS.

acoumb:
The PR will be merged after version _1.0.4_ has been released.

AndyButland (approved):

PR #58, opened 2022-11-16
Algolia Search Engine integration

acoumb:
This PR holds the integration with Algolia, a hosted search engine, consisting of two parts: _search implementation_ and _search analytics_. 
The integration's focus is the _search implementation_ component of Algolia. To ensure best compatibility with Algolia I am using the recommended _Algolia.Search_ library available [here](https://www.nuget.org/packages/Algolia.Search).
Library is used with following use cases:

- Index content: select content types with specific properties you want to index, pack them into a JSON payload and send the data package to Algolia.
- Search: for a specific index you can query the results and get matching hits. 
- Update index: when new content is submitted, if it is part of an Algolia index the index definition will get updated with new details.

The integration provides a custom dashboard with an UI inspired by the _Examine Search Management_ dashboard, where users can see existing indices with their indexed properties, can add new ones, edit existing ones or perform search against Algolia.

A different component of the integration is the handler watching when a new content gets published, and if a content type with properties is indexed in Algolia, the index definition will get updated.

We can schedule a demo session and go through all the use cases.


src/Umbraco.Cms.Integrations.Search.Algolia/Notifications/NewContentPublishedHandler.cs
```diff
@@ -0,0 +1,64 @@
...
+using Umbraco.Cms.Core.Events;
+using Umbraco.Cms.Core.Notifications;
+using Umbraco.Cms.Integrations.Search.Algolia.Builders;
+using Umbraco.Cms.Integrations.Search.Algolia.Migrations;
+using Umbraco.Cms.Integrations.Search.Algolia.Models;
+using Umbraco.Cms.Integrations.Search.Algolia.Services;
+
+namespace Umbraco.Cms.Integrations.Search.Algolia.Notifications
+{
+    public class NewContentPublishedHandler : INotificationHandler<ContentPublishedNotification>
```
AndyButland, line 14: I think - though you'll have to check - that these notification handlers can be asynchronous.  If so, you can make this one so, and the calls it makes to upstream components where you are currently having to do a `GetAwaiter().GetResult()` or similar, can be converted to `await`.

src/Umbraco.Cms.Integrations.Search.Algolia/Controllers/SearchController.cs
```diff
@@ -0,0 +1,124 @@
...
+            {
+                _scopeService.AddOrUpdate(new AlgoliaIndex
+                {
+                    Id = index.Id,
+                    Name = index.Name,
+                    SerializedData = JsonSerializer.Serialize(index.ContentData),
+                    Date = DateTime.Now
+                });
+
+                var payload = new List<Record>();
```
AndyButland, line 67: I'm still not sure why we need this part.  It could be very expensive to find all content by document type so I think this better removed and this operation is solely about saving index definitions to the database.

It would be the publish content events that should be used to actually populate the indexes.

That said, there might be value in having a separate option for fully re-indexing a particular index (i.e. deleting all the values in there, and re-creating using the code you have here).  But that should be a separate button, like on the examine dashboard, with copy indicating that it could take a long time.

src/Umbraco.Cms.Integrations.Search.Algolia/Notifications/NewContentPublishedHandler.cs
AndyButland, line 14: This is doing an update of existing content, so could be `NewContentPublishedHandler` isn't an ideal name (as it sounds like this is going to run only for newly created content).

AndyButland, line 14: We'll need a handler on deleting and unpublishing content too, so we can remove the records from the index.

src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaIndexService.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            try
+            {
+                var client = new SearchClient(_settings.ApplicationId, _settings.AdminApiKey);
+
+                var index = client.InitIndex(name);
+
+                index.SaveObjects(payload,  autoGenerateObjectId: false).Wait();
+
+                return string.Empty;
```
AndyButland, line 30: Not sure that we need this return value, as it seems to always be empty.  Perhaps these methods can just be `void` or `async Task`.

src/Umbraco.Cms.Integrations.Search.Algolia/Services/IScopeService.cs
```diff
@@ -0,0 +1,17 @@
+﻿
+namespace Umbraco.Cms.Integrations.Search.Algolia.Services
+{
+    public interface IScopeService<T>
```
AndyButland, line 4: This seems to be very generic - in the C# and English sense! - but it only seems to be used in one way.  Might be better just removing the generic parameter, and use the specific type.  And rename to something like `IAlgoliaIndexDefinitionStorage` (or similar)?

src/Umbraco.Cms.Integrations.Search.Algolia/Services/ScopeService.cs
```diff
@@ -0,0 +1,62 @@
...
+            using var scope = _scopeProvider.CreateScope();
+
+            return scope.Database.Single<AlgoliaIndex>("where Id = " + id);
+        }
+
+        public List<AlgoliaIndex> GetByContentTypeAlias(string alias)
+        {
+            using var scope = _scopeProvider.CreateScope();
+
+            return scope.Database.Fetch<AlgoliaIndex>("where SerializedData like '%" + alias + "%'");
```
AndyButland, line 46: This seems vulnerable to errors - could be that the property aliases and content type aliases match so you'd get false returns here.  Maybe you need to think about storing content types in a related table if you need to be able to get back all indexes that include a particular content type, or review if you really need this method.

AndyButland (commented):
Looks a good start and the pieces are in place.  Have a look over the comments and see what you think makes sense, and then let me know when we can review together (so I can see the UI too).  I'm available after 11 today.

src/Umbraco.Cms.Integrations.Search.Algolia/Handlers/ContentUnpublishedHandler.cs
```diff
@@ -0,0 +1,57 @@
+﻿using Microsoft.Extensions.Logging;
```
AndyButland, line 1: Looks like you could usefully introduce a base class for these handlers, that all three inherit from.  That would allow you to move common logic like iterating and filtering the indexes into one place.

src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaIndexService.cs
```diff
@@ -17,15 +17,24 @@ public AlgoliaIndexService(IOptions<AlgoliaSettings> options)
             _settings = options.Value;
         }
 
-        public string PushData(string name, List<Record> payload)
+        public string PushData(string name, List<Record> payload = null)
```
AndyButland, line 20: Can this be made `async` too, like the `DeleteData` method is?

src/Umbraco.Cms.Integrations.Search.Algolia/Umbraco.Cms.Integrations.Search.Algolia.csproj
```diff
@@ -17,6 +17,7 @@
 		<Authors>Umbraco HQ</Authors>
 		<Company>Umbraco</Company>
 		<PackageTags>Umbraco;Umbraco-Marketplace</PackageTags>
+		<PackageIcon>https://github.com/umbraco/Umbraco.Cms.Integrations/tree/main/src/Umbraco.Cms.Integrations.Search.Algolia/logo/algolia.png</PackageIcon>
```
AndyButland, line 20: You should check if this actually resolves to the image file.  You might need actually to click on the "Raw" link when you are viewing this page on GitHub, and use that as the URL.

src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaIndexService.cs
```diff
@@ -43,7 +52,29 @@ public async Task<string> UpdateData(string name, Record record)
...
                 var index = client.InitIndex(name);
 
-                await index.PartialUpdateObjectAsync(record);
+                var obj = index.GetObjects<Record>(new[] { record.ObjectID }).FirstOrDefault();
+                if (obj != null)
+                    await index.PartialUpdateObjectAsync(record);
+                else
+                    await index.SaveObjectAsync(record, autoGenerateObjectId: false);
+                
+                return string.Empty;
```
AndyButland, line 61: I see now why you are returning a string in that you check this in the caller to see if there has been an error.  But I don't see you return anything other than a string, so probably this could just be a void response.

Or, if you do want a result object, perhaps create a simple `Result` class, so it's a bit clearer what the response type is expected to contain.  Perhaps a simplified version of [this](https://github.com/umbraco/Umbraco.Marketplace/blob/main/src/PackagesRepository/PackagesRepository.Domain/Result.cs)?

src/Umbraco.Cms.Integrations.Search.Algolia/Controllers/SearchController.cs
```diff
@@ -52,61 +52,82 @@ public IActionResult GetIndices()
...
+
                 var payload = new List<Record>();
 
-                foreach (var item in index.ContentData)
+                var indexContentData = JsonSerializer.Deserialize<List<ContentData>>(index.SerializedData);
+
+                foreach (var contentDataItem in indexContentData)
                 {
-                    var contentItems = _umbracoHelper.ContentAtXPath($"//{item.ContentType}");
+                    var contentItems = _umbracoHelper.ContentAtXPath($"//{contentDataItem.ContentType}");
```
AndyButland, line 90: Might be better here to use `IContentService.GetPagedOfTypes()`.  You can do in a loop to not get massive amounts into memory in one go, and think this is a more Umbraco 9+ way to do things now there's no XML cache.

src/Umbraco.Cms.Integrations.Search.Algolia/Handlers/BaseContentHandler.cs
```diff
@@ -0,0 +1,64 @@
...
+            IAlgoliaIndexService indexService)
+        {
+            Logger = logger;
+
+            IndexStorage = indexStorage;
+
+            IndexService = indexService;
+        }
+
+        public async Task RebuildIndex(IEnumerable<IContent> entities, bool deleteIndexData = false)
```
AndyButland, line 31:
```suggestion
        protected async Task RebuildIndex(IEnumerable<IContent> entities, bool deleteIndexData = false)
```

src/Umbraco.Cms.Integrations.Search.Algolia/Handlers/BaseContentHandler.cs
```diff
@@ -0,0 +1,64 @@
...
+                        var record = new RecordBuilder()
+                           .BuildFromContent(entity, (p) => indexConfiguration.Properties.Any(q => q == p.Alias))
+                           .Build();
+
+                        var result = deleteIndexData
+                         ? await IndexService.DeleteData(index.Name, entity.Key.ToString())
+                         : await IndexService.UpdateData(index.Name, record);
+
+                        if (result.Failure)
+                            Logger.LogError($"Failed to delete data for Algolia index: {result}");
```
AndyButland, line 54:
```suggestion
                            Logger.LogError($"Failed to update the Algolia index: {result}");
```

AndyButland (approved):
Code-wise looks good now.  Will see again in the sprint demo later, and perhaps we'll have one more demo session next week when we can talk in more detail about the UI.

acoumb:
UI and controller updates.
An overview of the new dashboard UI can be viewed [here](https://drive.google.com/file/d/1nZ4t9P_g-bnmcMPsio9zUsd_WzH4Nof4/view?usp=sharing).

AndyButland (approved):
Looks great, approving, though just got a few more small UI points you might want to consider.

- The read more link should open in a new tab by default.
- It should be labelled "Read more about integrating Algolia search" (or similar, i.e. a bit clearer where you are going when you click).
- Let's label the "Add" button as "Add New Index Definition"
- When you click "Add" the order of the elements displayed is a little bit confusing, would suggest:
    - The "Manage Index" header should say "Create Index Definition" or "Edit Index Definition" depending on whether you are adding or editing.
    - Remove the "Content types" header
    - Move the "Please select the content types you want to index" message down below the "Document types" header and rename it to ""Please select the content types and properties you want to index"
- "Open" I think should say "Choose Properties" or similar
- When you list the indices on the dashboard main page, you are showing the property type alias rather than the name you are showing when managing the index.  Think the name is better here.
- Presumably we could show the property names rather than the aliases here too.

PR #59, opened 2022-11-21
Packages icons

acoumb:
Icons for packages.


AndyButland (changes requested):
Might be these are a bit big.  The NuGet spec states an image 128 x 128, but these are quite a bit wider.  Think it might be better to see if you can use a square(ish) icon, and then you can hopefully match the specified size.  Hopefully there's something appropriate for each of the integrations.

acoumb:
Re-uploaded versions of the icons:

- CommerceTools: 128x25
- ActiveCampaign: 128x13
- Dynamics: 128x48
- emerchantpay: 128x45
- Google Search Console: 128x46
- HubSpot: 128x37
- Semrush: 128x17
- Shopify: 128x37
- Zapier: 128x59

Most of the logos were picked from the official logo and marketing pack of the partner and include images that are difficult to bring to 128x128 without looking pixelated.

AndyButland (approved):
Looks like these will work fine - thanks for updating them.

AndyButland:
Though maybe ActiveCampaign is too small - could you perhaps just use the triangle?  E.g. [this one](https://ahkr.b-cdn.net/wp-content/uploads/2020/02/activecampaign-review-logo.png).

PR #60, opened 2022-12-02
Algolia integration using UUI library

acoumb:
This integration serves as an alternative to the previous dashboard design using the Umbraco 10 backoffice UI and API.

- UUI walkthrough: https://drive.google.com/file/d/1qQFROv9oaL5vcKGrMJyrbR2HT0bUOzDO/view?usp=sharing
- Umbraco 10 Backoffice UI: https://drive.google.com/file/d/1nZ4t9P_g-bnmcMPsio9zUsd_WzH4Nof4/view?usp=share_link

### DIFF
--- a/src/Umbraco.Cms.Integrations.Search.Algolia/App_Plugins/UmbracoCms.Integrations/Search/Algolia/css/algolia.css
+++ b/src/Umbraco.Cms.Integrations.Search.Algolia/App_Plugins/UmbracoCms.Integrations/Search/Algolia/css/algolia.css
@@ -1,5 +1,14 @@
-﻿.umb-content-grid {
-    display:grid;
-    grid-template-columns: repeat(3, 1fr);
-    gap:10px;
+﻿/* align child element in two columns: 1/3 and 2/3 */
+.alg-col-2 {
+    display: grid;
+    grid-template-columns: 25% 60%;
+    gap: 20px;
 }
+
+/* align child element in three equal columns */
+.alg-col-3 {
+    display: grid;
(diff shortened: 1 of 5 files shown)


AndyButland:
Looks pretty good so far... think it would be nice to use the UUI library here.  We can get used to it, and by open sourcing the project it'll be a good reference for others.

Just a couple of things that look a bit off to my eye:

 - The little "eye" icon you need to click on the open up the properties.  Wondering if that's very clear or necessary, could you not just click on the content item name and icon?
 - The delete property prompt hovers over the "Selected" button.
 - The icons for the document types are gone, would be nice if they could be retained.

Don't forget to reach out to the front-end team with issues if you can't find how to solve them with the library.  Even if they can't help, will be good feedback on issues being run into in use.

acoumb:
Additional changes to the Algolia integration and an updated UI [here](https://drive.google.com/file/d/1aLmS4tv7GBpLFNfjzjcr6p1uU0bH2Nan/view?usp=share_link).
Updates included with the last commit:

- `no-dirty-check` directive
- disable index name field in edit mode - the data is sent to Algolia based on the name of the index. Changing the name of an existing index will result in a new index created in Algolia and schema definition inconsistency.
- remove `selectable` attribute from `uui-ref-node` component displaying the list of content types. With the `selectable` attribute, when displaying the list of properties for the content type the focus from the ref node was removed, and there isn't a solution to mark the node as selected in the current UI library. Toggling the properties will be handled by  selecting the content type name.
- `tag` component changed to `primary` at the suggestion of Niels.

Issue #65, opened 2022-12-15
HubSpot API-key sunset


lars-erik:
HubSpot has deprecated the use of the "hapikey" querystring parameter since november 30th.  
The new "way" is to create a private app and send it's "secret" as a bearer token.  
PR incoming. 👼

acoumb:
Hi @lars-erik 
I have released version [2.0.0](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Crm.Hubspot/2.0.0) of HubSpot for Umbraco CMS containing your changes.
I will be closing this now, feel free to reach out if you have other input.
Thank you!

PR #66, opened 2022-12-15
Solves #65 HubSpot API-key sunset

lars-erik:
See #65

### DIFF
--- a/src/Umbraco.Cms.Integrations.Crm.Hubspot/Controllers/FormsController.cs
+++ b/src/Umbraco.Cms.Integrations.Crm.Hubspot/Controllers/FormsController.cs
@@ -100,7 +100,9 @@ namespace Umbraco.Cms.Integrations.Crm.Hubspot.Controllers
                 return new ResponseDto { IsValid = false };
             }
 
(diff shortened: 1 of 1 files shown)


acoumb:
Hi @lars-erik 
If my understanding is correct, with this PR, you enable users to use the access token of their private apps instead of the API key?

If so, wouldn't be best to use the access token of a private app as the third authentication method, maybe introduce a new setting in the site's configuration, and leave the API key as well? Although HubSpot specifies that using the API keys are a security risk, but I am thinking that maybe users who use it already might accept the risk and keep the configuration as it is.

Thank you

lars-erik:
Hey @acoumb,  
HubSpot has _discontinued_ support for a single API-key, I don't see the reason to leave the option in. :P  
https://developers.hubspot.com/changelog/upcoming-api-key-sunset

lars-erik:
Well, it's just a matter of time before it's completely turned off, and "fixing" requires like 5 clicks in your HubSpot portal after which you have a new key to copy into your configuration.  

The people who insist could just postpone upgrading this package as well?

acoumb:
Sounds good, merging it.

lars-erik:
Awesome, thanks! :)

PR #68, opened 2022-12-21
Feature/inriver integration

acoumb:
Current integration provides an entity picker for registered entities in Inriver, with a value converter providing a strongly typed model for rendering and a sample rendering view. The rendering view can be replaced by developers using the package with a view of their own.

Inriver is structured in two components:

- Control Center: users can set up their model, representing structured entity types, model that will be afterwards used in the Portal for defining the actual objects. In this interface, in the _Users_ section, the API base URL and API key can be found.
- Portal: frontend interface for defining entities based on the structure from the Control Center.

The _Inriver Entity Picker_ has a configuration section where you need to pick an entity type and the fields you want displayed in the property editor.

In the property editor, I query the entity IDs which have the entity type selected in the configuration, get entity summary for each ID, then retrieve the selected fields values.

A demonstration of the integration can be viewed [here](https://drive.google.com/file/d/1KTmFhi0R87fl9qQBENsfYEub3RfQJw9E/view?usp=share_link).


src/Umbraco.Cms.Integrations.sln
```diff
@@ -133,10 +139,17 @@ Global
 		{54A624E5-5321-4CC8-B74B-11ABF3605242}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{54A624E5-5321-4CC8-B74B-11ABF3605242}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{54A624E5-5321-4CC8-B74B-11ABF3605242}.Release|Any CPU.Build.0 = Release|Any CPU
+<<<<<<< HEAD
```
AndyButland, line 142: Looks like there's some merge conflict tags in here.

AndyButland (commented):
Looks great - very flexible and nice and easy to use.  Code looks fine.

A few comments on the UI...

Think you can make the content editor section a bit more editor friendly.  For example here:

- Should use names rather than IDs (so we remove the headings without spaces like "TaskStatus").
- Try to lose the term "entity" if you can.  The way I see this is that the developer will create one or more data types that will use the "InRiver Entity" property editor, but will be exposed to the editor as "Product Picker", "Task Picker" etc.  So it would be nice here to see the word "Products" or "Tasks" if that was possible.  Maybe that name is a configurable field on the property editor?
- Dates should be formatted (there's an angularjs filter you can use here). Hopefully there's some sort of type information you get back from the API to know whether something is a date rather than a string.
- Works nicely for a few items, but if you had hundreds we'd likely want paging and perhaps a text search filter.  If the API supports those things, would be good to set up (IIRC the CommerceTools product picker has an example here).

src/Umbraco.Cms.Integrations.PIM.Inriver/Models/FieldType.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Newtonsoft.Json;
+using System.Text.Json.Serialization;
+
+namespace Umbraco.Cms.Integrations.PIM.Inriver.Models
+{
+    public class FieldType
+    {
+        [JsonPropertyName("fieldTypeId")]
```
AndyButland, line 8: Do we need both of these?

AndyButland (approved):
Looks great.  One small point made but otherwise approving code-wise.

Let's just do one further UI run through tomorrow when you talk me through it.  Thanks.

acoumb:
Additional changes to the integration:

- hide first column in the products list table in case the selected entity types have no media resources (e.g. Task Picker vs Product Picker).
- update property editor configuration to allow the user to pick linked entities he wants to retrieve data for. The data from linked types will be available only through the view model, nothing is changing in the list of items from the property editor.
- code refactoring and bug fixing

src/Umbraco.Cms.Integrations.PIM.Inriver/App_Plugins/UmbracoCms.Integrations/PIM/Inriver/Render/InriverEntity.cshtml
```diff
@@ -1,15 +1,18 @@
...
     <p>@Model.DisplayDescription</p>
     @if (Model.Fields.Any())
     {
         <ul>
             @foreach (var field in Model.Fields)
             {
-                <li><b>@field.Key</b>: @field.Value</li>
+                <li>
+                    <b>@field.FieldTypeId</b>: @field.Value
+                    @field.Display
```
AndyButland, line 14: Maybe you should apply a bit more mark-up around this, so if someone wants to use this rendering component they could do so, and would have some hooks for styling.

AndyButland (approved):
The last round of changes look fine to me.

PR #71, opened 2023-01-06
Http codes validations and refactoring

acoumb:
Current PR extends the updates made for issue #65 by handling HTTP requests in a `try-catch` block. 

With HubSpot's update to remove API key based access, HubSpot users who don't want to use OAuth need to create a private app with `forms` scope and then use the private access token in the package. If the private app is not configured properly, a `Forbidden` status code will be returned by the API. In this case a specific error message notifying that the token has the incorrect permissions will be returned to the user.

For consistency, I have also updated the OAuth specific action of the controller to align with the one used by the API key.

Error messages have been moved server-side and the controller has been refactored.


AndyButland (approved):
Refactoring looks OK, but think we should add details of how to configure this to the readme page [here](https://github.com/umbraco/Umbraco.Cms.Integrations/tree/main/src/Umbraco.Cms.Integrations.Crm.Hubspot).  I'd hope that given it's a supported method HubSpot won't have any objections (as they did with the legacy API key).

Issue #69, opened 2023-01-02
Core projects/nuget packages


AndersBrohus:
Hi :)

Is there any chance we can get Core projects/nuget packages? :)

We are splitting our Umbraco models into a separated project, but are currently facing some issues because of the ActiveCampaign integration, since the only way to get the dll file into our models project is by referencing it from our build project - It works locally but not Azure 🙈

acoumb:
Hi @AndersBrohus 

I've released version [2.0.0](https://www.nuget.org/packages?q=umbraco.cms.integrations.crm.activecampaign) of the package with split versions.

I will close this for now, let me know if you have any input.

Thank you

PR #72, opened 2023-01-06
Split ActiveCampaign Core functionalities & support for Umbraco 11

acoumb:
Separate ActiveCampaign Core implementations per request in issue #69 and support for Umbraco 11.


AndyButland (approved):
Looks good.  Though you also need to bump the version, ideally to a v2 major version as the namespace of the models has changed.

PR #74, opened 2023-01-09
Fix issue with Algolia RecordBuilder where string arrays are pushed a…

Will-DiscoveryAgency:
Fixes issue in Algolia RecordBuilder where string arrays are pushed into the index as System.String[]


acoumb:
Hi @Will-DiscoveryAgency , Thank you for your contribution.

PR #75, opened 2023-01-10
V11/feature/algolia integration

acoumb:
Umbraco 11 support for Algolia integration.


src/Umbraco.Cms.Integrations.Search.Algolia/Umbraco.Cms.Integrations.Search.Algolia.csproj
```diff
@@ -22,11 +22,20 @@
 
 	<ItemGroup>
 		<PackageReference Include="Algolia.Search" Version="6.13.0" />
+	</ItemGroup>
+
+	<ItemGroup Condition="'$(TargetFramework)' == 'net60'">
```
AndyButland, line 27: Not sure you need the condition and change to the `TargetFrameworks` here.  Couldn't you just extend the single supported Umbraco version to `[10.3.0,12)`?

If you do that, I think - though you'll need to check - it'll work when referenced into an Umbraco 11 project.

Issue #70, opened 2023-01-05
Algolia Integration - Unable to add Document Types on Create Index Definition panel


ANT-GBR:
I'm using Umbraco 10.3.2 and I can't select Document Types on the Create Index Definition panel because the editor is not loading:

![image](https://user-images.githubusercontent.com/1932798/210815558-b6bbc798-ae5e-4558-9c49-e2b541032f6c.png)

Thanks
Tony

acoumb:
Hi Tony,

Could you tell me which version of Umbraco are you using?

ANT-GBR:
Hi Adrian,

I'm using Umbraco 10.3.2 and Umbraco.Cms.Integrations.Search.Algolia 1.0.0.

Thanks
Tony

d-gibbs:
Hey @acoumb, can you reproduce this issue in Umbraco v11? We have similar problem, although I don't see any console errors, the document types control just doesn't load.

Thanks

d-gibbs:
That does indeed seem to be the problem as if I create a content type at the root it appears in the control as expected:

![image](https://user-images.githubusercontent.com/59481722/211846439-6b18bdda-3eee-4678-8a6e-7681d258b758.png)

I think it needs to account for folders/nesting for it to work correctly.

Thanks

acoumb:
Thanks for clarifying this, I will take your input into consideration and address the matter.

acoumb:
Hi @d-gibbs ,
I've release version [1.1.1](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Search.Algolia/1.1.1) of the package with updates to content type retrieval.
I am closing this now, let me know if you have other input.

PR #76, opened 2023-01-12
Folder structured content types bug and version bump.

acoumb:
This PR fixes the issue reported [here](https://github.com/umbraco/Umbraco.Cms.Integrations/issues/70#issuecomment-1378950038), where content types structured in directories did not get retrieved when creating an index.

This was a consequence of the `contentTypeItem.parentId == -1` clause which was intended to skip composition elements.

The fix retrieves the entire list of content types, skips those that have the `isElement` flag `true`, and I use the `getWhereCompositionIsUsedInContentTypes` function of the API to check is the object is used as a composition.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Search.Algolia/App_Plugins/UmbracoCms.Integrations/Search/Algolia/js/algolia.service.js
+++ b/src/Umbraco.Cms.Integrations.Search.Algolia/App_Plugins/UmbracoCms.Integrations/Search/Algolia/js/algolia.service.js
@@ -1,17 +1,32 @@
 ﻿function algoliaService(contentTypeResource) {
     return {
         getContentTypes: function (callback) {
+
             contentTypeResource.getAll().then(function (data) {
-                callback(data.filter(item => item.parentId == -1 && !item.isElement).map((item) => {
-                    return {
-                        id: item.id,
-                        icon: item.icon,
-                        alias: item.alias,
-                        name: item.name,
-                        selected: false,
-                        allowRemove: false
+
+                var contentTypesArr = [];
+
+                for (let i = 0; i < data.length; i++) {
+                    
+                    if (data[i].isElement) continue;
+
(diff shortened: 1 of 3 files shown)


src/Umbraco.Cms.Integrations.Search.Algolia/App_Plugins/UmbracoCms.Integrations/Search/Algolia/js/algolia.service.js
```diff
@@ -1,17 +1,32 @@
...
+                                icon: data[i].icon,
+                                alias: data[i].alias,
+                                name: data[i].name,
+                                selected: false,
+                                allowRemove: false
+                            });
+                        }
+                    });
+
+                    if (data.length - 1 === i) {
```
AndyButland, line 26: Could you not just call `callback(contentTypesArr)` from outside the loop, instead of checking if you are on the last element from within the loop?

src/Umbraco.Cms.Integrations.Search.Algolia/App_Plugins/UmbracoCms.Integrations/Search/Algolia/js/algolia.service.js
```diff
@@ -1,17 +1,32 @@
...
-                        selected: false,
-                        allowRemove: false
+
+                var contentTypesArr = [];
+
+                for (let i = 0; i < data.length; i++) {
+                    
+                    if (data[i].isElement) continue;
+
+                    contentTypeResource.getWhereCompositionIsUsedInContentTypes(data[i].id).then(function (response) {
```
AndyButland, line 13: I'm wondering if it's actually correct to do this check.  In most cases it probably is, but you could have for example an "Article Page" and then a "Special Article Page", that composes the former and adds a few properties.  But they are both valid pages for editors to create.

So perhaps just filter our the elements and that'll do here.

AndyButland (approved):

Issue #77, opened 2023-01-20
Unable to connect to Dynamics 365 when creating Data Type


ben-umble:
Hi,

I'm trying to set up the Dynamics 365 Form Picker property editor for a client website but when I try to connect, it throws a generic error message:

`Dynamics Configuration: Error: An unexpected error occurred.`

I don't get any more information from the error unfortunately, here's the things that I've tried:

1. Tried with an admin user who has full permissions for the entire account
2. Ensured the DNS of the domain is valid
3. Triple-checked my configuration values in the `appSettings.json` file

I'm using the following versions of each platform/package:

1. Umbraco: `9.5.4`
2. Umbraco.Cms.Integrations.Crm.Dynamics: `1.1.2`

Please let me know if you need more information on this :)

Thanks,

Ben

mkalinov:
Hi guys, 
I have exactly same issue but in higher Umbraco version:
Umbraco 10.3.2
Forms 10.2.2
Deploy 10.1.2

I see my deployment logs, but there are no logs for the error though

acoumb:
Hi @ben-umble , @mkalinov ,
And thank you for bringing this to our attention. The key of the Azure App expired and authorization didn't go through.

I have renewed the key and you can use the package again, no other updates required on your part.

I will release though in the next days a minor patch with a handle of the _Unauthorized_ exception, to provide more clarity to the user.

I will be waiting for your confirmation, then I will close this.

Thank you.

mkalinov:
Good morning @acoumb 
Thanks for the update! I confirm it works now for me.

mkalinov:
@acoumb 
a question,
As I see u use custom types and Umbraco Integrations Azure App to make the picker work.
It doesn't looks reliable security wise, as Umbraco app has full access to customer data and we are fully dependent on u, as we just witnessed with key expiration.
We definitely do OAuth directly to crm api, can u please, refer us to any tutorial how to do a custom data type in Umbraco platform?
 

![image](https://user-images.githubusercontent.com/10182226/214093876-d525dd2a-aaa5-4d28-b481-a62c17166f58.png)

acoumb:
Hi,

Is [this](https://docs.umbraco.com/umbraco-cms/tutorials/creating-a-property-editor) what you're referring to?

[...]

PR #78, opened 2023-01-23
Handling HTTP exceptions

acoumb:
This PR is a follow up of issue #77 , where the secret key of the Azure App handling authentication expired, causing `401 Unauthorized` responses. The message displayed to the user was `Error: An unexpected error occurred.` without relevant information towards the error.

With this update, this scenario will be managed better in the future.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Crm.Dynamics/Controllers/FormsController.cs
+++ b/src/Umbraco.Cms.Integrations.Crm.Dynamics/Controllers/FormsController.cs
@@ -120,13 +120,10 @@ namespace Umbraco.Cms.Integrations.Crm.Dynamics.Controllers
 
                 return result;
 
-            } else
-            {
-                var errorResult = await response.Content.ReadAsStringAsync();
-                var errorDto = JsonConvert.DeserializeObject<ErrorDto>(errorResult);
+            var errorResult = await response.Content.ReadAsStringAsync();
+            var errorDto = JsonConvert.DeserializeObject<ErrorDto>(errorResult);
 
-                return "Error: " + errorDto.ErrorDescription;
-            }
+            return "Error: " + errorDto.ErrorDescription;
         }
 
         [HttpGet]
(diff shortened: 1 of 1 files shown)


src/Umbraco.Cms.Integrations.Crm.Dynamics/Controllers/FormsController.cs
```diff
@@ -96,20 +96,24 @@ public async Task<string> GetAccessToken([FromBody] OAuthRequestDto authRequestD
                 { "code", authRequestDto.Code }
             };
 
-            var requestMessage = new HttpRequestMessage
+            string responseContent = string.Empty;
+            try
```
AndyButland, line 100: This looks fine, but I'm a bit surprised it means you have to change the pattern used previously.  Could we not do a `if (response.IsSuccessStatusCode)` like we did before.

And then if that is `false`, examine: `if (response.StatusCode == HttpStatusCode.Unauthorized)`

acoumb: Updated

acoumb:
HTTP requests handling updated.

src/Umbraco.Cms.Integrations.Crm.Dynamics/Controllers/FormsController.cs
```diff
@@ -105,7 +105,7 @@ public async Task<string> GetAccessToken([FromBody] OAuthRequestDto authRequestD
             requestMessage.Headers.Add("service_name", AuthorizationService.Service);
 
             var response = await ClientFactory().SendAsync(requestMessage);
-            if (response.IsSuccessStatusCode)
+            if(response.IsSuccessStatusCode)
```
AndyButland, line 108:
```suggestion
            if (response.IsSuccessStatusCode)
```

src/Umbraco.Cms.Integrations.Crm.Dynamics/Controllers/FormsController.cs
```diff
@@ -119,17 +119,14 @@ public async Task<string> GetAccessToken([FromBody] OAuthRequestDto authRequestD
...
-            }
 
-            if (response.StatusCode == HttpStatusCode.BadRequest)
+            } else
             {
                 var errorResult = await response.Content.ReadAsStringAsync();
                 var errorDto = JsonConvert.DeserializeObject<ErrorDto>(errorResult);
 
                 return "Error: " + errorDto.ErrorDescription;
             }
```
AndyButland, line 129:
```suggestion
            var errorResult = await response.Content.ReadAsStringAsync();
            var errorDto = JsonConvert.DeserializeObject<ErrorDto>(errorResult);

            return "Error: " + errorDto.ErrorDescription;
```

AndyButland (approved):
Looks good.  I just left a couple of code styling suggestions you can accept or otherwise as you prefer.

PR #81, opened 2023-02-03
Fix variant issue

acoumb:
Current PR fixes a bug regarding content node variants. If a content node does not have variants, a `NULL` exception was thrown on the client when trying to retrieve the culture using `currentVariant.language.culture`.

To fix this, I am sending an empty string with the request, and pick the default language on the server.


AndyButland:
This looks like it'll fix the bug, but I wonder really if here it's better to look to retrieve the language of the current backoffice editor?  I.e. the one they have selected under their account in the _Users_ section.  That seems the best thing to use here, as we should be showing the information in the language most suited to the user in the backoffice.

acoumb:
Updated to take the logged in backoffice user culture into consideration.

AndyButland (approved):
Looks good to me,

Issue #85, opened 2023-02-08
Integrations.Search.Algolia - Include document URL with indexed item


d-gibbs:
Hi,

I'm using `Umbraco.Cms.Integrations.Search.Algolia 1.1.1` and in the process of implementing site search.

I have found that there is no ability to select a document URL when choosing which fields to index for a given type.

For example - I've created a basic document type which has a single property `Title`:

![image](https://user-images.githubusercontent.com/59481722/217574042-7598ea1e-a162-4a7d-859f-331986873bc6.png)

And the index definition includes the single property as expected:

![image](https://user-images.githubusercontent.com/59481722/217574208-6cc41bb0-0ae3-40ad-b136-6063d3555b46.png)

But you get nothing else other than this. I feel to be useful, the index item at a minimum should include a URL to the page. Without this, we'd need to perform lookups on the items coming back in order to be able to give the user a link to the search result.

Could the indexer include the document URL as default OR can you provide the option to pick default umbraco fields (as well custom properties) to include e.g. content type, culture, url, id etc... ?

Thank you

chris-evansnz:
@acoumb in addition to URL it would be good to get the node Name as well

tony-gledhill:
Please can we include the node created and updated dates.

acoumb:
Hi @d-gibbs , @chris-evansnz , @tony-gledhill ,

Version [1.2.0](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Search.Algolia/1.2.0) of the package contains the additional properties.

Thank you,
Adrian

tony-gledhill:
Thanks @acoumb - I'm installing now ! 👍 H5YR

acoumb:
Version [1.3.0](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Search.Algolia/1.3.0) is now released, with an updated indexing and [documentation](https://docs.umbraco.com/umbraco-dxp/integrations/algolia).

Closing this for now, but feel free to reach out if you have other input.

Issue #84, opened 2023-02-08
Integrations.Search.Algolia - New documents are not being pushed/indexed on save & publish


d-gibbs:
Hi,

I'm using `Umbraco.Cms.Integrations.Search.Algolia 1.1.1` and in the process of implementing site search. I have a basic index which includes two types with a number of fields defined to include against them.

Example:

- Article page
  - Title
  - Body content
  - Image
- Content page
  - Title
  - Body content

I'm finding that creating a new page of either of these document types does not trigger the index to be rebuilt and the new page does not show up as an additional index item when checking [algolia.com](https://algolia.com).

I have also found that changes to existing documents, that are already included in the index, do not get pushed to algolia either e.g. a change to the `Title` or `Body content` field above.

The only way to get these to show up is if I head to `Settings > Algolia Search Management` and manually rebuild the index.

The [readme](https://github.com/umbraco/Umbraco.Cms.Integrations/tree/main/src/Umbraco.Cms.Integrations.Search.Algolia) suggests that:

> The actual content payload is pushed to Algolia for indices created in Umbraco on two scenarios:
> 
> - from the list of indices, the Build action is triggered, resulting in all content of specific document types to be sent as JSON to Algolia.
> - **_using a ContentPublishedNotification handler which will check the list of indices for the specific document type, and will update a matching Algolia object_**

So it seems the plugin should be handling this but something isn't working quite right.

I have checked log viewer and can see the following error raised upon saving a page:

![image](https://user-images.githubusercontent.com/59481722/217572498-c251eae5-1ac6-44dc-9fa2-4fbe6efec09c.png)

If there's any further info you need from me just let me know.

Thank you

acoumb:
Hi @d-gibbs ,

I could not replicate this issue so I tried to retrace your steps as shown in [this](https://www.youtube.com/watch?v=mCMB4bjmGXA&ab_channel=UmbracoDXP) video. 

Could you check my use case and see if in your case there is something different, so I can identify the issue?

Thank you,
Adrian

d-gibbs:
Hi @acoumb 

Thanks for getting back to me. I've attached a screen recording:
https://www.awesomescreenshot.com/video/14772203?key=0da69266847238b73188bde050fd586a

In the video I perform the following actions:

- Show that the document type `Article` is included as an allowed document type
- Show the algolia index in the CMS and Algolia - `GlobalSiteSearch`, note the record count sits at **25**
[...]

Issue #82, opened 2023-02-03
Integrations.Search.Algolia - Index rebuild doesn't handle complex types e.g. MediaWithCrops


d-gibbs:
Hi,

I'm using `Umbraco.Cms.Integrations.Search.Algolia 1.1.1` and in the process of implementing site search. I have set up a basic search index in the backend and have selected a number of fields to be included, one of which is an image field which Umbraco treats as a `MediaWithCrops` object.

I'm finding that the indexer is treating these fields as simple strings where ideally we'd like to get back a path to the media item.

Example:

![image](https://user-images.githubusercontent.com/59481722/216632954-e42894c2-effa-477f-a7f3-431029386d89.png)

Is this something the plugin is capable of out of the box? If not, does the plugin have a way of intercepting the rebuild event on certain document types so that we can add this logic ourselves?

Thank you

acoumb:
Hi @d-gibbs ,

Could you tell me what editor are you using to get this payload in Algolia?

My `Image Media Picker` that is based on `Umbraco.MediaPicker3` indexed the result as show in [this](https://www.youtube.com/watch?v=mCMB4bjmGXA&ab_channel=UmbracoDXP) video, with this JSON value: `[{"key":"b0567635-3f07-4b32-b636-55b965a47081","mediaKey":"55514845-b8bd-487c-b370-9724852fd6bb"}]`.

Thanks,
Adrian

d-gibbs:
Hi @acoumb

That document type is using a control based on the `Umbraco.MediaPicker3`:

![image](https://user-images.githubusercontent.com/59481722/218494676-170450f4-e065-4207-ba64-7c18e3bdaf87.png)

![image](https://user-images.githubusercontent.com/59481722/218495143-22891559-4f9c-4c64-99ff-ffc72fe3e53f.png)

Field in algolia:

![image](https://user-images.githubusercontent.com/59481722/218495297-bddc0ddb-60e5-466b-b833-6e10fd4630a8.png)

FWIW, we are using models builder which is set to `SourceCodeManual`.

Thank you

acoumb:
Hi @d-gibbs ,

Please find the updates in version [1.2.0](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Search.Algolia/1.2.0) of the package.

Thank you,
Adrian

acoumb:
[...]

PR #86, opened 2023-02-15
Fix null reference bug & parsing of complex types

acoumb:
The updates in this PR handle the following issues reported:

- #84 - updated code that returned a `NULL` reference.
- #82 
- #85 

I have introduced an `Umbraco Library` that currently holds only parsing components, receiving as input an `IContent -> IProperty` or `IPublishedContent -> IPublishedProperty`, and depending on the value type will returned a parsed value.

`RecordBuilder` is now split in two other builders:
- `ContentRecordBuilder` - content node is published
- `PublishedContentRecordBuilder` - index build is triggered from the dashboard

In the record payload that gets serialized and sent to Algolia, I have included properties for `Name` and `Document URL`, and properties that have variants can be retrieved from the `Data` dictionary using this template `[alias]-[culture]`.


src/Umbraco.Cms.Integrations.Search.Algolia/Builders/ContentRecordBuilder.cs
```diff
@@ -0,0 +1,69 @@
...
+            _parserService = parserService;
+        }
+
+        public ContentRecordBuilder BuildFromContent(IContent content, Func<IProperty, bool> filter = null)
+        {
+            _record.ObjectID = content.Key.ToString();
+
+            _record.Id = content.Id;
+            _record.Name = content.Name;
+            _record.Url = _urlProvider.GetUrl(content.Id);
```
tony-gledhill, line 32: Would it be possible to add the created and updated unix dates here please?

src/Umbraco.Cms.Integrations.Search.Algolia/Builders/PublishedContentRecordBuilder.cs
```diff
@@ -0,0 +1,67 @@
...
+            _parserService = parserService;
+        }
+
+        public PublishedContentRecordBuilder BuildFromContent(IPublishedContent publishedContent, Func<IPublishedProperty, bool> filter = null)
+        {
+            _record.ObjectID = publishedContent.Key.ToString();
+
+            _record.Id = publishedContent.Id;
+            _record.Name = publishedContent.Name;
+            _record.Url = _urlProvider.GetUrl(publishedContent.Id);
```
tony-gledhill, line 31: Would it be possible to add the created and updated unix dates here please?

tcmorris:
Hey @acoumb @AndyButland would you be able to provide a timeline on when this might be merged and the package updated?

acoumb:
> Hey @acoumb @AndyButland would you be able to provide a timeline on when this might be merged and the package updated?

Hi @tcmorris , 
I am striving for the beginning of next week.

acoumb:
With this new update the Algolia indexing undergoes some changes, as it is now based on the `IDataEditor.PropertyIndexValueFactory` property from the CMS. The `GetIndexValues` method is designed to return values in a search engine friendly manner.
- [`ContentBuilder`](https://github.com/umbraco/Umbraco.Cms.Integrations/blob/feature/algolia-extend-indexing/src/Umbraco.Cms.Integrations.Search.Algolia/Builders/ContentRecordBuilder.cs) is responsible for creating the record object that will be pushed to _Algolia_.
- [`AlgoliaSearchPropertyIndexValueFactory`](https://github.com/umbraco/Umbraco.Cms.Integrations/blob/feature/algolia-extend-indexing/src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaSearchPropertyIndexValueFactory.cs) will return the property value using `PropertyIndexValueFactory`, either the default value or a converted one for a complex type (in this case, the `Umbraco.MediaPicker3` property editor).

To assign methods for handling a specific property editor, the [`Converters`](https://github.com/umbraco/Umbraco.Cms.Integrations/blob/456e832fcaa50f1636906c78b1f4da71a46b7604/src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaSearchPropertyIndexValueFactory.cs#L26) will be used, by passing the proper values in the constructor.

If one might want to use their own implementation, then they can inherit `AlgoliaSearchPropertyIndexValueFactory`, override the `GetValue` method and pass their own converters (e.g. [this](https://github.com/umbraco/Umbraco.Cms.Integrations/blob/feature/algolia-extend-indexing/src/Umbraco.Cms.Integrations.Search.Algolia/Services/ExtendedAlgoliaSearchPropertyIndexValueFactory.cs) sample implementation).

src/Umbraco.Cms.Integrations.Library/Parsing/MediaWithCropsParser.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    public class MediaWithCropsParser : BaseParser<MediaWithCrops>
+    {
+        public override string Parse(MediaWithCrops input)
+        {
+            if(input.LocalCrops != null && input.LocalCrops.Src != null)
+                return input.LocalCrops.Src;
+
+            return string.Empty;
+        }
```
AndyButland, line 15:
```suggestion
        public override string Parse(MediaWithCrops input) => input.LocalCrops?.Src ?? string.Empty;
```

src/Umbraco.Cms.Integrations.Library/Parsing/MediaWithCropsParser.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        public override string Parse(List<MediaWithCrops> input)
+        {
+            if(input.Count() > 0)
+            {
+                var srcArr = input
+                    .Where(p => p.LocalCrops != null && p.LocalCrops.Src != null)
+                    .Select(p => p.LocalCrops.Src)
+                    .ToArray();
```
AndyButland, line 24:
```suggestion
                var srcArr = Parse(input)
                    .Where(p => !string.IsNullOrEmpty(p)))
                    .ToArray();
```

src/Umbraco.Cms.Integrations.Library/Parsing/MediaWithCropsParser.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            if(input.LocalCrops != null && input.LocalCrops.Src != null)
+                return input.LocalCrops.Src;
+
+            return string.Empty;
+        }
+
+        public override string Parse(List<MediaWithCrops> input)
+        {
+            if(input.Count() > 0)
```
AndyButland, line 19:
```suggestion
            if (input.Any())
```

src/Umbraco.Cms.Integrations.Library/Entities/LibraryMedia.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Text.Json.Serialization;
```
AndyButland, line 1: Just to validate - do we think this makes sense as a separate package with re-use across various integrations?  Or is it better to fold back into Algolia if it's only used there?  I'm not sure... but just wanted to you to consider before we commit to this.

src/Umbraco.Cms.Integrations.Search.Algolia/Models/LibraryMedia.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Text.Json.Serialization;
```
AndyButland, line 1: This seems to be duplicated in the library project.

src/Umbraco.Cms.Integrations.Search.Algolia/Services/ExtendedAlgoliaSearchPropertyIndexValueFactory.cs
```diff
@@ -0,0 +1,34 @@
+﻿
+using System.Text.Json;
```
AndyButland, line 2: I think this class (and it's commented out registration) should probably be in the docs - for now, in the readme - rather than in the code.

src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaSearchPropertyIndexValueFactory.cs
```diff
@@ -0,0 +1,90 @@
+﻿using System.Text.Json;
+
+using Umbraco.Cms.Core.Models;
+using Umbraco.Cms.Core.Services;
+
+namespace Umbraco.Cms.Integrations.Search.Algolia.Services
+{
+    public class AlgoliaSearchPropertyIndexValueFactory : IAlgoliaSearchPropertyIndexValueFactory
```
AndyButland, line 8: I've not quite followed how this class works with the `ParserService` you've also defined in the library project.  It seems to be doing a similar thing and has some duplicated code from there.

Are they two versions of work doing similar things?  If so, think we should try to combine so there's only one class responsible got "getting property values for the index".

acoumb:
Recent updates no longer use the _Library_, handling property value is managed in `AlgoliaSearchPropertyIndexValueFactory`.

src/Umbraco.Cms.Integrations.Search.Algolia/Models/LibraryMedia.cs
AndyButland: I see you've removed that now so it's no longer duplicated, but is `LibraryMedia` now the right name?

AndyButland (approved):

PR #87, opened 2023-02-21
Feature/aprimo integration

acoumb:
Current PR provides an integration with [Aprimo](https://www.aprimo.com/platform/digital-asset-management), a Digital Asset Management platform, for Umbraco CMS versions 10.3+, providing a custom media picker used as a property editor, a value converter with a strongly typed model to render the selected asset.

A video demonstration can be viewed [here](https://youtu.be/nsblP2P6V2E). As Aprimo does not support Firefox and Internet Explorer, my demo toggles between Firefox and Edge to provide a behavioral insight into working with Aprimo Cloud.

An editor can manage DAM assets in two ways:

1. Using the Aprimo API and picking a media item from property editor list.
2. Using the [Content Selector](https://developers.aprimo.com/digital-asset-management/aprimo-integration-tools/aprimo-content-selector/) and selecting the asset directly from Aprimo Cloud.

For handling authorization, I am using the [Authorization Code Flow with Proof Key for Code Exchange (PKCE)](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow-with-proof-key-for-code-exchange-pkce), by generating a code verifier and a code challenge and store them into the Umbraco database.

The code challenge will be sent with the first request for authorization and the Aprimo Authorization Server saves it. With the received code the code verifier will be sent in the second request to retrieve the access token and the refresh token.

The scopes used are `api` and `offline_access`, meaning that we are receiving a refresh token with the access token, and use it to renew the access token in case a `401 Unauthorized` code is received. This is why in several places I check the value of the `IsAuthorized` flag, and request the server to renew the access token.


src/Umbraco.Cms.Integrations.Crm.Dynamics/Configuration/OAuthSettings.cs
```diff
@@ -0,0 +1,20 @@
+﻿
+namespace Umbraco.Cms.Integrations.Crm.Dynamics.Configuration
```
AndyButland, line 2: Is this included in error?  Looks like it's related to Dynamics and presumably the other work planned to allow customers to bring their own app for authentication.  If so, let's keep that as a separate update.

src/Umbraco.Cms.Integrations.DAM.Aprimo/App_Plugins/UmbracoCms.Integrations/DAM/Aprimo/Render/AprimoAsset.cshtml
```diff
@@ -0,0 +1,9 @@
+﻿@inherits Umbraco.Cms.Web.Common.Views.UmbracoViewPage<Umbraco.Cms.Integrations.DAM.Aprimo.Models.ViewModels.AprimoAssetViewModel>
+
+<div>
```
AndyButland, line 3: If the implementer was going to use this directly, they could benefit from some hooks for CSS - i.e. class name on the surrounding div, and removal of the inline styles.

src/Umbraco.Cms.Integrations.DAM.Aprimo/Services/AprimoService.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        public async Task<AprimoResponse<Record>> SearchRecord(Guid id)
+        {
+            var client = _httpClientFactory.CreateClient(Constants.AprimoClient);
+
+            var oauthConfiguration = _oauthConfigurationStorage.Get();
+            if (oauthConfiguration == null) return AprimoResponse<Record>.Fail(Constants.ErrorResources.Unauthorized, false);
+
+            client.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("Umbraco.Cms.Integrations.DAM.Aprimo", "1.0.0"));
```
AndyButland, line 30: Looks like some of this code is repeated in the other `SearchRecords` method - so seems like you could create some private helper methods to DRY this up.

AndyButland, line 30: Also, headers that are always the same can be set as `DefaultRequestHeaders` where the named client is setup, so they don't need to be set on every request.

Things that change per request should probably be on the headers of the request rather than the default headers of the client.

src/Umbraco.Cms.Integrations.DAM.Aprimo/Services/IAprimoService.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Umbraco.Cms.Integrations.DAM.Aprimo.Models;
+
+namespace Umbraco.Cms.Integrations.DAM.Aprimo.Services
+{
+    public interface IAprimoService
+    {
+        Task<AprimoResponse<SearchedRecordsPaged>> SearchRecords(string page);
+
+        Task<AprimoResponse<Record>> SearchRecord(Guid id);
```
AndyButland, line 9: This could perhaps be better named as `GetRecordById`?  Just as it's not really a search.

src/Umbraco.Cms.Integrations.DAM.Aprimo/App_Plugins/UmbracoCms.Integrations/DAM/Aprimo/Render/AprimoAsset.cshtml
acoumb: Removed

src/Umbraco.Cms.Integrations.Crm.Dynamics/Configuration/OAuthSettings.cs
acoumb: Removed

src/Umbraco.Cms.Integrations.DAM.Aprimo/App_Plugins/UmbracoCms.Integrations/DAM/Aprimo/Render/AprimoAsset.cshtml
acoumb: Updated

src/Umbraco.Cms.Integrations.DAM.Aprimo/Services/AprimoService.cs
acoumb: Updated

acoumb: Updated

src/Umbraco.Cms.Integrations.DAM.Aprimo/Services/IAprimoService.cs
acoumb: Updated

acoumb:
Additional changes included in the PR based on feedback and demos:

- allow users to choose the desired workflow for picking content: using the _Aprimo Content Selector_ or the _Content API_. To do so, the `Use Content Selector` option from the data type configuration is to be used. If `Use content selector` option if true, then the property editor overlay will not be shown, the `Add` will open the _Content selector_ directly. If false, then you will be able to retrieve the media assets through the API, and filter them by a search term or go to different pages.
- authorization process is detached from the `OAuthProxy` and it is self hosted, exposing a controller endpoint (_https://localhost:44324/umbraco/api/authorization/oauth_) for handling the response from the `Aprimo Authorization Server`. 
If the authorization code is received, I will request the access tokens and then inject through `ContentResult` JavaScript code that sends a message to the parent window.
- regarding the content preview, I have raised an [issue](https://github.com/umbraco/Umbraco.UI/issues/430) with the frontend team as the image does not fit the card.

The recorded demo: [https://youtu.be/ShqDRvtA_JI](https://youtu.be/ShqDRvtA_JI)

src/Umbraco.Cms.Integrations.DAM.Aprimo/App_Plugins/UmbracoCms.Integrations/DAM/Aprimo/Render/AprimoAsset.cshtml
```diff
@@ -0,0 +1,11 @@
+﻿@inherits Umbraco.Cms.Web.Common.Views.UmbracoViewPage<Umbraco.Cms.Integrations.DAM.Aprimo.Models.ViewModels.AprimoAssetViewModel>
+
+<link href="~/App_Plugins/UmbracoCms.Integrations/DAM/Aprimo/css/aprimo.css" rel="stylesheet"/>
+
+<div class="aprimo-container">
+    @if (!string.IsNullOrEmpty(Model.Thumbnail))
+    {
+        <img class="aprimo-thumbnail" src="@Model.Thumbnail" alt="@Model.Title" />
+    }
+    <h3 class="aprimo-title">@Model.Title</h3>
```
AndyButland, line 10: I reckon this should probably be a `div`, as we can't know really what heading level is appropriate semantically here.  And the implementer can style this as a heading if they wanted to based on the CSS classes you have added.

src/Umbraco.Cms.Integrations.DAM.Aprimo/App_Plugins/UmbracoCms.Integrations/DAM/Aprimo/css/aprimo.css
```diff
@@ -0,0 +1,25 @@
+﻿.aprimo-editor-container {
```
AndyButland, line 1: It's perhaps best to not provide this, such that the implementer can style how they want.  As it stands any custom styles would conflict with once we've provided here.

src/Umbraco.Cms.Integrations.DAM.Aprimo/Models/ViewModels/AprimoAssetViewModel.cs
```diff
@@ -0,0 +1,14 @@
+﻿
+namespace Umbraco.Cms.Integrations.DAM.Aprimo.Models.ViewModels
+{
+    public class AprimoAssetViewModel
+    {
+        public string Id { get; set; } = string.Empty;
+
+        public string Title { get; set; } = string.Empty;
+
+        public string Thumbnail { get; set; } = string.Empty;
```
AndyButland, line 10: How would you get different sized images for rendering on the front-end?  Currently I can only see rendering the thumbnail, but it could be we'd want the full-size version or different crops.  Is any of that available?

AndyButland (commented):
Looking good.  Just had a comment on the data type configuration from the video...

In the property editor, think it would be good to have some text indicating what "not" using the content selector would mean - and on the other hand, there's probably too much text shown when you do select to "Use content selector" (e.g. all the detail about ID or CDN mode, which doesn't seem useful unless there's a reason and option to choose between them).

I'd suggest something like:

_"Assets will be retrieved using an Umbraco interface communicating with the Aprimo API. This works with all browsers supported by Umbraco."_

_"Assets will be retrieved using a richer Aprimo user interface, but is only available when using the Chrome or Edge browser.  Read more."_

---

Other than inline comments, the only other concern was the authorization endpoint path - `umbraco/api/authorization/oauth` - which seems a bit generic (and hence perhaps could clash with something else from another package or implementation.  So think you need to make this path a little more specific to this Aprimo package.

src/Umbraco.Cms.Integrations.DAM.Aprimo/App_Plugins/UmbracoCms.Integrations/DAM/Aprimo/Render/AprimoAsset.cshtml
acoumb: Updated, rendering also the cropped items.

acoumb:
Additional updates:
- retrieve cropped items
- retrieve additional fields and their values based on language
- updates to rendering component
- updates to models based on the additional data retrieved from the API.

src/Umbraco.Cms.Integrations.DAM.Aprimo/Models/ViewModels/AprimoAssetViewModel.cs
```diff
@@ -9,6 +9,14 @@ public class AprimoAssetViewModel
 
         public string Thumbnail { get; set; } = string.Empty;
 
-        public string Tag { get; set; } = string.Empty;
+        public IEnumerable<AprimoCropItemViewModel> CropItemsVM { get; set; }
```
AndyButland, line 12: Not sure we need the `VM` suffix here, as it doesn't seem to be consistent with other view model fields.  Why not just `CropItems` (or `Crops`) and `Fields`?

src/Umbraco.Cms.Integrations.DAM.Aprimo/App_Plugins/UmbracoCms.Integrations/DAM/Aprimo/Render/AprimoAsset.cshtml
```diff
@@ -1,11 +1,53 @@
 ﻿@inherits Umbraco.Cms.Web.Common.Views.UmbracoViewPage<Umbraco.Cms.Integrations.DAM.Aprimo.Models.ViewModels.AprimoAssetViewModel>
 
-<link href="~/App_Plugins/UmbracoCms.Integrations/DAM/Aprimo/css/aprimo.css" rel="stylesheet"/>
+<style type="text/css">
```
AndyButland, line 3: As per other comment, I'm wondering if we should provide this.  Of course it makes it look better to have some styling, but I'm wondering what happens if someone wants to update it.  They can't override this in a stylesheet, and they can't edit the file.

So maybe we provide in the package just the rendering view - with CSS classes.

And provide in documentation a sample set of styling that makes the rendering look reasonable.

The implementer can then either use our partial view and apply styles.  Or, if that's not enough, create their own rendering component using ours as a reference.

src/Umbraco.Cms.Integrations.DAM.Aprimo/App_Plugins/UmbracoCms.Integrations/DAM/Aprimo/Render/AprimoAsset.cshtml
```diff
@@ -1,11 +1,53 @@
...
+    <div class="aprimo-asset-row">
+        <h3>Crop Items</h3>
+        <br/>
+        <div class="aprimo-crops-row">
+            @foreach (var cropItem in Model.CropItemsVM)
+            {
+                <div class="column">
+                    @if (!string.IsNullOrEmpty(cropItem.PublicLink))
+                    {
+                        <img src="@cropItem.PublicLink" class="thumbnail" />
```
AndyButland, line 45: Probably should render the `width` and `height` for the `img` tag using the `ResizeWidth` and `ResizeHeight` properties.

Similarly `Label` could be used for the `alt` attribute.

src/Umbraco.Cms.Integrations.DAM.Aprimo/AprimoComposer.cs
```diff
@@ -39,7 +39,8 @@ public void Compose(IUmbracoBuilder builder)
                     client.DefaultRequestHeaders.Add("API-VERSION", "1");
                     client.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("Umbraco.Cms.Integrations.DAM.Aprimo", "1.0.0"));
                     client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
-                    client.DefaultRequestHeaders.Add("select-record", "title,tag,thumbnail");
+                    client.DefaultRequestHeaders.Add("select-record", "title,thumbnail,fields,masterfilelatestversion");
```
AndyButland, line 42: My guess would be that these two aren't needed for all request, just the one where you get an asset by Id?

If so, they probably shouldn't be default headers sent on all requests, rather added to the `Headers` collection for just the request that needs it.

src/Umbraco.Cms.Integrations.DAM.Aprimo/Services/IAprimoService.cs
```diff
@@ -4,7 +4,9 @@ namespace Umbraco.Cms.Integrations.DAM.Aprimo.Services
 {
     public interface IAprimoService
     {
-        Task<AprimoResponse<SearchedRecordsPaged>> SearchRecords(string page, string searchTerm);
+        Task<AprimoResponse<SearchItemsPaged<Language>>> GetLanguages();
```
AndyButland, line 7: Maybe by adding async and sync methods here - e.g. `GetLanguages()` and `GetLanguagesAsync()`, `SearchRecords()` and `SearchRecordsAsync()` - you could make it a bit easier to use for calling code to select the right one?  Then you just have one place to remember to do all the `ConfigureAwait` /  `GetAwaiter().GetResult()` pieces.

src/Umbraco.Cms.Integrations.DAM.Aprimo/Umbraco.Cms.Integrations.DAM.Aprimo.csproj
```diff
@@ -9,7 +9,7 @@
 	<PropertyGroup>
 		<PackageId>Umbraco.Cms.Integrations.DAM.Aprimo</PackageId>
 		<Title>Umbraco CMS Integrations: DAM - Aprimo</Title>
-		<Description>An extension for Umbraco CMS providing a media picker for digital assets managed in Aprimo workspace..</Description>
+		<Description>An extension for Umbraco CMS providing a media picker for digital assets managed in Aprimo workspace.</Description>
```
AndyButland, line 12:
```suggestion
		<Description>An extension for Umbraco CMS providing a media picker for digital assets managed in an Aprimo workspace.</Description>
```

acoumb:
Additional updates added to the PR:
- `async`/`sync` methods for services
- update client headers to retrieve the public CDN URLs used for rendering the media asset: original and cropped items
- updated the rendering component using the Aprimo item details as reference, and display the crop name, extension, size and render the image from the Public CDN of Aprimo resized
image
- code refactoring

src/Umbraco.Cms.Integrations.DAM.Aprimo/App_Plugins/UmbracoCms.Integrations/DAM/Aprimo/Render/AprimoAsset.cshtml
```diff
@@ -31,23 +10,36 @@
...
+                    @if (!string.IsNullOrEmpty(cropItem.Url))
                     {
-                        <img src="@cropItem.PublicLink" class="thumbnail" />
+                        <img src="@cropItem.Url" class="thumbnail" alt="@cropItem.Name" />
                     }
-                    <p><b>Label:</b> @cropItem.Label</p>
-                    <p><b>Size:</b> @cropItem.ResizeWidth x @cropItem.ResizeHeight</p>
+                    <h3>@cropItem.Name</h3>
+                    <p>
+                        @cropItem.Extension - @($"{cropItem.ResizeWidth} x {cropItem.ResizeHeight} px")
```
AndyButland, line 39: Should we use these to add `width` and `height` attributes to the image?

src/Umbraco.Cms.Integrations.DAM.Aprimo/Services/AprimoService.cs
```diff
@@ -57,6 +62,10 @@ public async Task<AprimoResponse<Record>> GetRecordById(Guid id)
 
             client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", accessToken);
 
+            client.DefaultRequestHeaders.Add("select-record", "title,thumbnail,fields,masterfilelatestversion");
```
AndyButland, line 65: Consider moving these headers to the request's headers collection rather than the client's default request headers.  I'm not sure, but could be how you are doing it these headers would be used in another request unless overridden.  There's a few examples of this in this class.

AndyButland (approved):

src/Umbraco.Cms.Integrations.DAM.Aprimo/App_Plugins/UmbracoCms.Integrations/DAM/Aprimo/Render/AprimoAsset.cshtml
acoumb: The URL points to the resized image, so adding the attributes will not make any difference.
For example this media asset: https://p1.sb.aprimocdn.net/partner1/b648ad4f-743e-4af3-8363-afaf00d4a59d/Wireless%20headset%20listening%20to%20music%20304328931_Social.jpg, which already comes as 1920 x 1080 px.

AndyButland: May still be good practice to add them (at least it was in 2005 but I may be out of date now!).  Just as it means the browser reserves space for the image before loading it, so the screen doesn't jump when it has.

Issue #97, opened 2023-04-11
Integrations.Search.Algolia - unable to provide search API key


tcmorris:
Hi,

Within the configuration I was hoping to be able to define the [search API key](https://www.algolia.com/doc/guides/security/api-keys/#search-only-api-key) that Algolia suggest should be used for searching as well as any front end integration (e.g. using the [Autocomplete](https://www.algolia.com/doc/guides/building-search-ui/ui-and-ux-patterns/autocomplete/js/) control) since we don't want to be sharing our admin API key anywhere client side.

However, only the admin API key can be added in the configuration. I can see that there is a [wrapper for performing searches](https://github.com/umbraco/Umbraco.Cms.Integrations/blob/main/src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaSearchService.cs#L22) against the Algolia client so it would make sense to include.

It would be useful to update the documentation with a reference to the Algolia guides for the API keys as well, to ensure that they get used appropriately.

acoumb:
Hi @tcmorris ,

I've released version [1.3.1](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Search.Algolia/1.3.1) of the integration that contains the Search API key as well, based on your suggestions.

Also, details on this update have been included in the [Umbraco Docs](https://docs.umbraco.com/umbraco-dxp/integrations/algolia) as well.

Thank you for your input,
Adrian

PR #98, opened 2023-04-21
Include Search API key

acoumb:
Current PR introduces the additional Search API key as part of the Algolia settings, for a safe use to perform searches on the frontend.

With this update, I am addressing issue #97 and the input provided there.


AndyButland (approved):
Looks fine to me.  Please also don't forget to document this with a PR to the docs site.

Issue #105, opened 2023-05-23
Integrations.Search.Algolia - Unable to index RTE null value


rspencer-aimwa:
Hello,

Our website is on-
`Umbraco.Cms` version `10.5.1`
`Umbraco.Cms.Integrations.Search.Algolia` version `1.3.1` (temporarily reverted to `1.1.1`)

The issue is happening on `Umbraco.Cms.Integrations.Search.Algolia` `1.3.1`, but we have temporarily reverted to `1.1.1`, which lets us manually build index

Our content isn't getting indexed on 'save and publish', or on manual index
When trying to manually build index on any of our Algolia Indices we get the following error:
![image](https://github.com/umbraco/Umbraco.Cms.Integrations/assets/134269042/b376f44c-4c5b-44e9-9f4c-e48216b9ee94)

On doing some error testing, I was able to pinpoint it to when one of the 'Page Properties' marked on one of the 'Document Types' in the 'Index Definition' is an RTE without a value, then it throws the error.

The error is getting caught at `RebuildIndex` in `AlgoliaContentCacheRefresherHandler`.
Which is getting sent from `GetValue` in `AlgoliaSearchPropertyIndexValueFactory`.
Where a `KeyValuePair` that would have the key of the 'property alias' is `null`.
This seems to be happening because of the `yield break` in `RichTextPropertyIndexValueFactory` from `RichTextPropertyEditor` (in `Umbraco.Infrastructure`).
Giving `KeyValuePair` null values instead of being null itself, so it passes by the null check in  `GetValue` from `AlgoliaSearchPropertyIndexValueFactory`.
I can see a couple of possible fixes either in Umbraco `RichTextPropertyIndexValueFactory` or the `AlgoliaSearchPropertyIndexValueFactory` null check.

Let me know if there is any other details you need from me

Thank you

acoumb:
Hi @rspencer-aimwa ,

Thank you for reporting this issue, I have released a new [version](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Search.Algolia/1.3.2) of the package with a fix for this.

Please let me know how this works for you now.

Thanks,
Adrian

rspencer-aimwa:
Hello @acoumb ,
That looks to have done the job
Thank you!

PR #106, opened 2023-05-24
Fix indexing for empty RTEs

acoumb:
Current PR fixes issue #105 , when indexing an empty RTE throws an exception. Also, instead of returning the default value of the `KeyValuePair` type, I still add the selected indexed property to the record payload, but with an empty value.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaSearchPropertyIndexValueFactory.cs
+++ b/src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaSearchPropertyIndexValueFactory.cs
@@ -34,7 +34,7 @@ namespace Umbraco.Cms.Integrations.Search.Algolia.Services
 
             var indexValues = dataType.Editor.PropertyIndexValueFactory.GetIndexValues(property, culture, string.Empty, true);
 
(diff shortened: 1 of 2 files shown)


src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaSearchPropertyIndexValueFactory.cs
```diff
@@ -34,7 +34,7 @@ public AlgoliaSearchPropertyIndexValueFactory(IDataTypeService dataTypeService,
 
             var indexValues = dataType.Editor.PropertyIndexValueFactory.GetIndexValues(property, culture, string.Empty, true);
 
-            if(indexValues == null) return default;
+            if (indexValues == null || !indexValues.Any()) return new KeyValuePair<string, string>(property.Alias, string.Empty);
 
             var indexValue = indexValues.FirstOrDefault();
```
AndyButland, line 39:
```suggestion
            var indexValue = indexValues.First();
```

Just be be more explicit that you have already checked there is a value and you won't be returning a default/null value.

AndyButland (approved):
Looks fine but had one suggestion just to make the code a bit more explicit.

PR #109, opened 2023-05-30
Feature/v12 support

acoumb:
Current PR follows the testing of the integrations against Umbraco CMS v12.0.0-rc2.
Updates included:
- Zapier & Dynamics: moved from deprecated `IScopeProvider` to `ICoreScopeProvider` for .NET 6 and 7 which was throwing an exception when trying to execute the migration.
- updated versions and packages references versioning


AndyButland (commented):
This looks OK, but can you try using `IScopeProvider` found in `Umbraco.Cms.Infrastructure.Scoping` for Umbraco 10, 11 and 12?  I think that may work in all cases.  There was some rework in 10 that introduced this and the one in this namespace was the backward compatible one.  Not sure, but you may find that works in all and hence you can simplify a bit.

acoumb:
Updated Zapier and Dynamics integrations to use `IScopeProvider`, for .NET5 from `Umbraco.Cms.Core.Scoping` and for .NET6 and above, use from `Umbraco.Cms.Infrastructure.Scoping`.

AndyButland (approved):
That looks better, thanks.

src/Umbraco.Cms.Integrations.Crm.Dynamics/Umbraco.Cms.Integrations.Crm.Dynamics.csproj
```diff
@@ -23,9 +23,21 @@
...
+	</ItemGroup>
+
+	<ItemGroup Condition="'$(TargetFramework)' == 'net60'">
+		<PackageReference Include="Umbraco.Cms.Web.Website" version="[10.0.0,11)" />
+		<PackageReference Include="Umbraco.Cms.Web.BackOffice" version="[10.0.0,11)" />
+		<PackageReference Include="Umbraco.Cms.Web.Common" version="[10.0.0,11)" />
+	</ItemGroup>
+
+	<ItemGroup Condition="'$(TargetFramework)' == 'net70'">
+		<PackageReference Include="Umbraco.Cms.Web.Website" version="[11.0.0,12]" />
```
nul800sebastiaan, line 38: @acoumb I think this is a mistake, it should have a round brace behind v12, instead of a straight bracket. 
I have a report from a user that can't upgrade to v12 because of this.  I've tested it as well and I get the error below.
Must have worked in your tests as it was a pre-release, so maybe that doesn't count as v12 yet? 😅 

image

acoumb: This has been addressed with a new patch release. 
Thank you

PR #110, opened 2023-06-06
Allow properties to be in the varyByCulture state and without if the …

gardarthorsteins:
If node was set to VaryByCulture the builder will look at the properties to all be VaryByCulture.

But we could have some properties with out the VaryByCulture settings.

We now look for VaryByCulture on each property instead of only checking if the node is VaryByCulture.


acoumb:
Hi @gardarthorsteins,

Thank you for your PR and your support. I have included the updates in version [1.3.3](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Search.Algolia/1.3.3) of the package.

Thanks,
Adrian

PR #114, opened 2023-06-29
Fixed XPATH error when multiple root nodes

redmorello:
We noticed when using this package in our installation that has multiple root nodes, that the ContentAtXPath lookup on line 119 was failing. Once I'd changed this line to retrieve all nodes of contenttype, then the indexing issue was resolved.


acoumb (approved):
Update looks good.

acoumb:
Hi @redmorello ,
Thank you for your support, I will include the update with the next minor release of the package.

acoumb:
Update available with version [1.4.0](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Search.Algolia/1.4.0)

PR #116, opened 2023-06-30
Update package.manifest in ActiveCampaign integration

AndersBrohus:
We kept having issues when Umbraco was running in production mode, and after a while we figured out we need to have the following line inside package.manifest

"bundleOptions": "None" 

And then i removed one line which was not included


acoumb (approved):
Updates approved.
Thank you

acoumb:
Hi @AndersBrohus ,

Thank you for your support, the updates have been included with version [2.0.2](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Crm.ActiveCampaign/2.0.2) of the package.

Regards,
Adrian

PR #117, opened 2023-07-11
Show loaders while oauth and loading forms is not completed

EelmanE:
Show loaders while oauth and loading forms is not completed. 
Also remove LoadForms call from promise to prevent double loading of forms list.


acoumb (approved):
Looks good. Thank you.

PR #118, opened 2023-07-11
Toggle API key validation between CMS and Forms package settings

acoumb:
Current PR is a result of the tests made for [this](https://github.com/umbraco/Umbraco.Forms.Integrations/issues/86) Zapier Forms issue, where an exception is shown when trying to retrieve the list of forms.

Trying to replicate the use case, I have identified an issue when trying to authenticate an user, regarding the base URL, which could be saved with a trailing slash `/`, causing the next API requests to fail.

When a user sets up a new Zap and configures the trigger, he needs to
- pick an app and select an event:
image

- set up an account and authenticate it against the Umbraco website through an API key, or credentials:
image

In the account configuration window, if the base URL was not inputted properly, any `404` responses would have been handled incorrectly because the Zapier App response handling code looks only for a `false` content: 
image
image

Additionally, the authorization service from the CMS package would lookup the API key in the CMS settings, instead of picking them from Forms settings.

Applied fixes:
- Handle the response based on the HTTP status code and then check the content of the response: 
image
- When validating user access, if CMS settings are missing, move and check the settings for Forms.

Currently the Zapier App changes under version 2.1.0 are in private mode, once the package update will be ready, I will make them public.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Automation.Zapier/Configuration/AppSettings.cs
+++ b/src/Umbraco.Cms.Integrations.Automation.Zapier/Configuration/AppSettings.cs
@@ -1,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Umbraco.Cms.Integrations.Automation.Zapier.Configuration
+{
+    public abstract class AppSettings
+    {
+        public string UserGroupAlias { get; set; }
+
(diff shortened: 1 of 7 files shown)


AndyButland (approved):
Looks reasonable to me.  There are a few things involved though so maybe it would be good to merge this in and make a pre-release available to the customer.  You could put it up on the MyGet feed and let them have access to that to try to see if solves their issue.  If so, can then go out as a patch release proper.

Issue #124, opened 2023-07-31
Adding a RichText field to index, then update index crashes Umbraco


danbord:
Adding a richText field to Index and building index crashes Umbraco

using Umbraco 12

danbord:
Same issue with ImageCrop field

danbord:
Some finds.  I've been able to fix the Access Violation crash.

In AlgoliaSearchPropertyIndexValueFactory.GetValue, it was crashing on the line :
`var indexValues = dataType.Editor.PropertyIndexValueFactory.GetIndexValues(property, culture, string.Empty, true)`

Seems that overload is deprecated, and suggest that we use the overload that passes the availableCultures.  Passing the availableCultures fixes the Crash.  But we end up with nothing being indexed for that field.

Here's my fix:

need to inject localizationService, then:
```
var allLanguages = _localizationService.GetAllLanguages().Select(l => l.IsoCode);
var indexValues = dataType.Editor.PropertyIndexValueFactory.GetIndexValues(property, culture, string.Empty, true, allLanguages);
```

acoumb:
Hi @danbord ,

After considering the implications of this change to different versions of the CMS, as detailed below, we think that making this change will impact developers using versions of Umbraco `10.3` to `10.5.1`, as well as `11.0.0` to `11.3.1`, and the solution would be to update the Umbraco CMS to the version containing the fix. In your case, updating to `12.1.0` will fix the issue.
![image](https://github.com/umbraco/Umbraco.Cms.Integrations/assets/95346674/f083181a-8116-4837-96fc-71ea8c509a10)

Closing the issue, thank you for your feedback,
Adrian

PR #125, opened 2023-08-04
Fix deprecated method invocation

acoumb:
Current PR fixes issue #124 caused by a method being deprecated in recent versions of Umbraco.

The fix uses the issue reporter's recommendations, but to avoid compile errors, I had to modify the minimum required version of `Umbraco.Cms.Web.BackOffice` to `10.6.0`.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaSearchPropertyIndexValueFactory.cs
+++ b/src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaSearchPropertyIndexValueFactory.cs
@@ -2,6 +2,7 @@
 
 using Umbraco.Cms.Core.Models;
 using Umbraco.Cms.Core.Services;
+using Umbraco.Extensions;
 
 namespace Umbraco.Cms.Integrations.Search.Algolia.Services
 {
@@ -11,11 +12,15 @@ namespace Umbraco.Cms.Integrations.Search.Algolia.Services
 
         private readonly IMediaService _mediaService;
 
-        public AlgoliaSearchPropertyIndexValueFactory(IDataTypeService dataTypeService, IMediaService mediaService)
+        private readonly IEnumerable<string> _availableLanguages;
+
+        public AlgoliaSearchPropertyIndexValueFactory(IDataTypeService dataTypeService, IMediaService mediaService, ILocalizationService localizationService)
         {
             _dataTypeService = dataTypeService;
 
             _mediaService = mediaService;
+
+            _availableLanguages = localizationService.GetAllLanguages().Select(p => p.IsoCode);
 
             Converters = new Dictionary<string, Func<KeyValuePair<string, IEnumerable<object>>, string>>
             {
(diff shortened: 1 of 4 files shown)


src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaSearchPropertyIndexValueFactory.cs
```diff
@@ -11,12 +12,16 @@ public class AlgoliaSearchPropertyIndexValueFactory : IAlgoliaSearchPropertyInde
...
-        public AlgoliaSearchPropertyIndexValueFactory(IDataTypeService dataTypeService, IMediaService mediaService)
+        private readonly IEnumerable<string> _availableLanguages;
+
+        public AlgoliaSearchPropertyIndexValueFactory(IDataTypeService dataTypeService, IMediaService mediaService, ILocalizationService localizationService)
         {
             _dataTypeService = dataTypeService;
 
             _mediaService = mediaService;
 
+            _availableLanguages = localizationService.GetAllLanguages().Select(p => p.IsoCode);
```
AndyButland, line 23: Is this service registered a singleton?  If so, this wouldn't get any updates to the languages created in Umbraco until a restart.  And if so. might mean this needs to be saved as `_localizationService` and the `GetAllLanguages()` method called at the time of indexing.

src/Umbraco.Cms.Integrations.Search.Algolia/Umbraco.Cms.Integrations.Search.Algolia.csproj
```diff
@@ -24,7 +24,7 @@
 	<ItemGroup>
 		<PackageReference Include="Algolia.Search" Version="6.13.0" />
 		<PackageReference Include="Umbraco.Cms.Web.Website" version="[10.3.0,)" />
-		<PackageReference Include="Umbraco.Cms.Web.BackOffice" version="[10.3.0,)" />
+		<PackageReference Include="Umbraco.Cms.Web.BackOffice" version="[10.6.0,)" />
```
AndyButland, line 27: If updating the reference - which is fine - you should update the other `Umbraco.Cms` ones to to be on a consistent version.

src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaSearchPropertyIndexValueFactory.cs
```diff
@@ -29,11 +34,11 @@ public AlgoliaSearchPropertyIndexValueFactory(IDataTypeService dataTypeService,
         {
             var dataType = _dataTypeService.GetByEditorAlias(property.PropertyType.PropertyEditorAlias)
                 .FirstOrDefault(p => p.Id == property.PropertyType.DataTypeId);
-
+         
             if (dataType == null) return default;
 
-            var indexValues = dataType.Editor.PropertyIndexValueFactory.GetIndexValues(property, culture, string.Empty, true);
-
+            var indexValues = dataType.Editor.PropertyIndexValueFactory.GetIndexValues(property, culture, string.Empty, true, _availableLanguages);
```
AndyButland, line 40: Makes sense to use the non-deprecated method, but I don't understand the fix I'm afraid.  Looking in the CMS code, at `RichTextPropertyIndexValueFactory`, which was mentioned in the issue as being one that causes the problem, I can't see where the `availableLanguages` parameter is used.  And as such why passing in `Enumerable.Empty<string>` isn't fine here.

Probably I'm missing something - but if we do need to move the call to retrieve to all languages here, it's potentially an expensive thing to do so wanted to make sure it is necessary.

acoumb:
Considering the impact of this change on other versions of the CMS, there is no point in making this change as patch releases to the CMS will fix the issue. Closing the PR.
image

Issue #133, opened 2023-09-26
Umbraco.Cms.Integrations.Crm.Hubspot plugin breaks razor precompilation


digbyswift:
We have a Umbraco v10.7.0 project with Umbraco.Cms.Integrations.Crm.Hubspot installed and we wanted to move to using view precompilation. We have view precompilation set up on a range of other projects but in this particular project, inclusion of this plugin breaks the precompilation process with the following error.

> CSC: Warning CS8785 : Generator 'RazorSourceGenerator' failed to generate source. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'ArgumentException' with message 'The hintName 'App_Plugins_UmbracoCms_Integrations_Crm_Hubspot_Render_HubspotForm_cshtml.g.cs' of the added source file must be unique within a generator. (Parameter 'hintName')'

Our umbraco project has a dependency on a second project that itself has a reference to the Umbraco.Cms.Integrations.Crm.Hubspot plugin. This reference is for the purposes of accessing the `HubSpotFormViewModel`.

Because both projects reference the plugin, the `Render/*.cshtml` files are duplicated in each project. It is this duplication which is preventing the view compilation.

For now we are working round this by having the plugin reference only in the second, dependent project but the solution would presumably to have a second plugin package that contained only the core classes, e.g. Umbraco.Cms.Integrations.Crm.Hubspot.Core

acoumb:
Hi @digbyswift ,

I have released a [new major version](https://www.nuget.org/packages?q=umbraco.cms.integrations.crm.hubspot) for the HubSpot package with separation in `*.Core` project.

Regards,
Adrian

digbyswift:
Brilliant, thank you so much.

PR #134, opened 2023-09-28
Feature/hubspot core

acoumb:
Following issue #133 , I've split the HubSpot functionalities into a separate `*.Core` project.


AndyButland (approved):
This looks good to me and a sensible update.  I think it needs to be a major version bump though, as it's a breaking change due to all the namespace changes.  So should be 4.0.0.

I think you can probably leave updating the V8 version of the package.  The person that requested this is on Umbraco 10, and I think it's reasonable we move on with updates to 10+ unless there's a good reason to update the 8 version too.

It also makes versioning easier.  If you do want to update the 8 version, then that should be a major version too, which I guess means that goes to 4.0.0 and this to 5.0.0.

PR #135, opened 2023-09-28
Separate Hubspot functionality into Core project

acoumb:
Following issue #133 , I've split the HubSpot functionalities into a separate `*.Core` project.


AndyButland (commented):
As discussed in https://github.com/umbraco/Umbraco.Cms.Integrations/pull/134, I'd be minded to leave the 8 version as is and progress with this change only for 10+.

acoumb:
Makes sense, leaving this as it is.

PR #139, opened 2023-10-18
disable build button while loading and show loading text

gardarthorsteins:
This will fix so the build button will be disabled if we are waiting for the job to finish.

Also <umb-loading-indicator> does not do anything so at least I added the loading... text into it so its visible.


acoumb (approved):

acoumb:
Looks good @gardarthorsteins . Thank you for your contribution.

PR #140, opened 2023-10-18
Feature/performance update

gardarthorsteins:
We had 7000 nodes that we were trying to index into Angolia with out a success.

After looking into the build code we could see that it took 2-3sec to build each record and there for would have taken around 5 hours to finish.

We were working locally and connecting to an azure database. This would make it slower and we understand that it would have been faster on the production machine.

For each record the code was calling GetById into content service and also for each property on that content it would call _dataTypeService.GetByEditorAlias. This would create huge amount of db connection that slows down the process immensly.

We could not see any reason to use content services or any services that connects to the db for this function as we only want published nodes into the Angolia Index.

This update will change it so we only use IPublishedProperty and IPublishedContent instead of the IContent and IProperty.

After this change we could see huge performance difference. each record took 10-20ms instead of 2000-2500ms.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Search.Algolia/Controllers/SearchController.cs
+++ b/src/Umbraco.Cms.Integrations.Search.Algolia/Controllers/SearchController.cs
@@ -1,9 +1,10 @@
-﻿using Algolia.Search.Models.Search;
+﻿using System.Diagnostics;
+using Algolia.Search.Models.Search;
 
 using Microsoft.AspNetCore.Mvc;
 
 using System.Text.Json;
-
+using Microsoft.Extensions.Logging;
 using Umbraco.Cms.Core.Routing;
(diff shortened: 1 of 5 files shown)


src/Umbraco.Cms.Integrations.Search.Algolia/Controllers/SearchController.cs
```diff
@@ -122,18 +128,27 @@ public async Task<IActionResult> BuildIndex([FromBody] IndexConfiguration indexC
 
                 foreach (var contentDataItem in indexContentData)
                 {
+                    var sw = new Stopwatch();
```
acoumb, line 131: I will be removing the stopwatch before releasing the new version, as I think it fits better for testing purposes.

acoumb (approved):
Looks good, thank you for your contribution.

PR #145, opened 2023-10-31
Bugfix/zapier auth fix

acoumb:
Current PR adds a fix to the API key based validation flow, based on this Forms package issue reported on Zapier's developer platform:
image
image
image

Test case to replicate the issue:
1. Have API key for both CMS and Forms
2. Create a new Zap -> user is authenticated with CMS key
3. Select `New form submitted` trigger
4. Pull forms step returns an empty list, because the saved key from step 2. is CMS, and Forms package expects one for Forms.

To fix the issue, I am checking both CMS and Forms keys and validate both.


src/Umbraco.Cms.Integrations.Automation.Zapier/Umbraco.Cms.Integrations.Automation.Zapier.csproj
```diff
@@ -9,7 +9,7 @@
 		<RepositoryUrl>https://github.com/umbraco/Umbraco.Cms.Integrations</RepositoryUrl>
 		<PackageProjectUrl>https://github.com/umbraco/Umbraco.Cms.Integrations/blob/main/src/Umbraco.Cms.Integrations.Automation.Zapier</PackageProjectUrl>
 		<Product>Umbraco.Cms.Integrations.Automation.Zapier</Product>
-		<Version>1.2.0</Version>
+		<Version>1.3.0</Version>
```
AndyButland, line 12: Would think this could just be a patch release.

AndyButland (approved):

Issue #149, opened 2023-11-07
Algolia indexes only root (Homepage)


diger74:
Umbraco version: 12.3.0
Algolia integration package version: 1.5.0

For some reason the content of my custom content type pages (Case Study Details Page) is not indexed into Algolia, the log says "indexed 0 items", although there are 6 pages in the CMS tree.

What I have discovered in the code while investigating this issue:

```
    using (UmbracoContextReference contextReference = this._umbracoContextFactory.EnsureUmbracoContext())
    {
        // contentType here is not null, all is ok
        var contentType = contextReference.UmbracoContext.Content.GetContentType("caseStudyDetailsPage");

        // this returns 0 items, this method is used by Algolia integration package
        var byContentType = contextReference.UmbracoContext.Content.GetByContentType(contentType)
            .ToList();

        // this returns 6 case study items, although this is exactly the same code as in GetByContentType() method
        var fromRootByContentType = contextReference.UmbracoContext.Content.GetAtRoot()
            .SelectMany(x => x.DescendantsOrSelf(_variationContextAccessor!))
            .Where(x => x.ContentType.Id == contentType.Id)
            .ToList();
    }
```

For some reason unknown .GetByContentType() method, which is used inside Algolia integration package, returns 0 items and therefore there's nothing to send to Algolia.

However, when I call .GetAtRoot().SelectMany(..).Where(..), which is exactly the same code as in .GetByContentType() method, it returns all my 6 case study items with no issues.

---
_This item has been added to our backlog AB#34636_

diger74:
@acoumb a bit more context on this issue: this happens when I create a new index having some content in the CMS already.
When the index is created (originally with 0 items), and I start adding or amending new pages (e.g. case studies in my example) - those items start appearing in Algolia once published in Umbraco.

I suppose the main problem is with the way the content is fetched for the full index rebuild, once individual items publish notifications are synched into Algolia with no issues.

diger74:
Hi @acoumb,

This is likely linked with this PR https://github.com/umbraco/Umbraco.Cms.Integrations/pull/153
Apologize, I didn't mention in my original post, that my content types used in the test were composed of other content types. As mentioned in the referenced PR this might be the issue.

acoumb:
Hi @diger74 ,

[...]

PR #153, opened 2023-11-14
Fix for retrieval of content by type when rebuilding Algolia index and blocks list value getter

piotrkrajewskicn:
The current code that is getting a list of content by type is not (always) working, this is an Umbraco bug which causes the following call: `var contentItems = ctx.UmbracoContext.Content.GetByContentType(contentType)` to return an empty list for some content types (as far as I know, it is connected to types that are composed of other types). In order to avoid this problem, I switched to using `GetPagedOfType` from `ContentService`. 

As a side effect, since the new code makes use of `IContent` and not `IPublishedContent`, we can remove the extra code that was added in this PR: https://github.com/umbraco/Umbraco.Cms.Integrations/pull/140 by @gardarthorsteins 

It is also worth mentioning that the removed code had an extra bug in `AlgoliaSearchPropertyIndexValueFactory`'s `KeyValuePair<string, string> GetValue(IPublishedProperty property, string culture)`: it used `property.GetSourceValue(culture)` in order to get a value to index but as contrary to `GetIndexValues` from `IPropertyIndexValueFactory`, it did not return correct data to index for the data type of `Umbraco.BlockList` which is another reason to get rid of the new method of `GetValue` for `IPublishedProperty`. Additionally, the method introduced some code duplication.

As the mentioned PR from @gardarthorsteins attempted to fix performance issues, the new version is handling them by making use of an injected `PropertyEditorCollection` in `AlgoliaSearchPropertyIndexValueFactory` so there is no need to make multiple calls for data types and their editors from `IDataTypeService`. I tested the performance on indexing over 9000 nodes which is taking on average 12 seconds on my local machine so around 1.3 millisecond for entry

### DIFF
--- a/src/Umbraco.Cms.Integrations.Search.Algolia/Builders/ContentRecordBuilder.cs
+++ b/src/Umbraco.Cms.Integrations.Search.Algolia/Builders/ContentRecordBuilder.cs
@@ -26,56 +26,7 @@ namespace Umbraco.Cms.Integrations.Search.Algolia.Builders
 
             _algoliaSearchPropertyIndexValueFactory = algoliaSearchPropertyIndexValueFactory;
         }
-        public ContentRecordBuilder BuildFromContent(IPublishedContent content, Func<IPublishedProperty, bool> filter = null)
-        {
-            _record.ObjectID = content.Key.ToString();
 
-            _record.Id = content.Id;
-            _record.Name = content.Name;
-
-            _record.CreateDate = content.CreateDate.ToString();
-            _record.CreatorName = content.CreatorName();
-            _record.UpdateDate = content.UpdateDate.ToString();
-            _record.WriterName = content.WriterName();
-
-            _record.TemplateId = content.TemplateId.HasValue ? content.TemplateId.Value : -1;
-            _record.Level = content.Level;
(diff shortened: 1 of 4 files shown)


piotrkrajewskicn:
Hi @AndyButland , would you mind checking out this PR? The other PR (https://github.com/umbraco/Umbraco.Cms.Integrations/pull/154) that you just accepted makes use of the code here in this PR so it'd great to have it merged, it also fixes this issue: https://github.com/umbraco/Umbraco.Cms.Integrations/issues/149

acoumb:
Hi @piotrkrajewskicn ,

I have prepared a new version of Algolia to be shipped out today, that contains some refactoring of the converters, and I have manually added some of your changes. 

I will update here once the new version is on NuGet.

Thank you,
Adrian

acoumb:
Version [2.0.0](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Search.Algolia/2.0.0) has been released, and it contains a refactoring for the converters using collection builder (more details [here](https://github.com/umbraco/Umbraco.Cms.Integrations/pull/154)).

I have manually picked changes you've recommended.

Thank you for your contribution.

Adrian

Issue #151, opened 2023-11-09
Algolia Search - Indexing of non-string values


geann:
Sometimes we need to index non-string fields in Algolia. For example, we have a MultiNodeTreePicker category field and we want to index it as an array of names or GUIDs. Currently we have to serialize this data because the method `GetValue()` of `AlgoliaSearchPropertyIndexValueFactory` can return only string values. 
The problem with this approach is that it is then stored as a text value in Algolia rather than an object:

![image](https://github.com/umbraco/Umbraco.Cms.Integrations/assets/4091048/7eb7d7f8-7f83-4a58-bee3-5affaa8067b1)

Even though full text search by categories works ok, there is an issue with facets as Algolia doesn't parse serialised objects and instead treats them as text:
![image](https://github.com/umbraco/Umbraco.Cms.Integrations/assets/4091048/d51fedd0-523c-4287-9d31-2b3c7b4d7f98)

I tracked usage of string values down to the class [Record.cs](https://github.com/umbraco/Umbraco.Cms.Integrations/blob/main/src/Umbraco.Cms.Integrations.Search.Algolia/Models/Record.cs) where the `Data` property is of type `Dictionary<string, string>`. Do you think this property can be changed to `Dictionary<string, object>` to cater for non-string values?

---
_This item has been added to our backlog AB#34719_

thenexus00:
It is a json string that you need so that comes into the Algolia search objects which can process in there. I have internal fixes and I did do a pull request so things like Multi tree node picker renders out the name. You then do work in your Algolia to handle that.

acoumb:
Hi @geann ,

Thank you for the reported issue and your [PR](https://github.com/umbraco/Umbraco.Cms.Integrations/pull/152). 

I've done some testing of different use cases and as a result we would need to prepare in the coming weeks a new version of the package that will provide additional converters for sending data to Algolia, and options for developers to manage the converters themselves, by adding new or removing existing ones.

I will update here with the progress.

Thank you,
Adrian

geann:
That's great, thank you @acoumb!

acoumb:
Hi @geann ,

We have released version [2.0.0](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Search.Algolia/2.0.0) with changes to the way converters work (more details [here](https://github.com/umbraco/Umbraco.Cms.Integrations/pull/154)). 

Please update to the new version and let me know how it goes.

Thank you,
Adrian

Issue #150, opened 2023-11-07
Algolia Search - Application crashes with Stack Overflow exception when indexing media picker properties


geann:
Whenever I select a property of type `Umbraco.MediaPicker3` for an Algolia index and try to save changes on an indexable page, the application crashes with the following error:

```
Stack overflow.
Repeat 23771 times:
--------------------------------
   at Umbraco.Cms.Core.PropertyEditors.NoopPropertyIndexValueFactory.GetIndexValues(Umbraco.Cms.Core.Models.IProperty, System.String, System.String, Boolean)
--------------------------------
   at Umbraco.Cms.Integrations.Search.Algolia.Services.AlgoliaSearchPropertyIndexValueFactory.GetValue(Umbraco.Cms.Core.Models.IProperty, System.String)
   at Umbraco.Cms.Integrations.Search.Algolia.Builders.ContentRecordBuilder.BuildFromContent(Umbraco.Cms.Core.Models.IContent, System.Func`2<Umbraco.Cms.Core.Models.IProperty,Boolean>)
   at Umbraco.Cms.Integrations.Search.Algolia.Handlers.AlgoliaContentCacheRefresherHandler+<RebuildIndex>d__10.MoveNext()
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[[Umbraco.Cms.Integrations.Search.Algolia.Handlers.AlgoliaContentCacheRefresherHandler+<RebuildIndex>d__10, Umbraco.Cms.Integrations.Search.Algolia, Version=1.5.0.0, Culture=neutral, PublicKeyToken=null]](<RebuildIndex>d__10 ByRef)
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start[[Umbraco.Cms.Integrations.Search.Algolia.Handlers.AlgoliaContentCacheRefresherHandler+<RebuildIndex>d__10, Umbraco.Cms.Integrations.Search.Algolia, Version=1.5.0.0, Culture=neutral, PublicKeyToken=null]](<RebuildIndex>d__10 ByRef)
   at Umbraco.Cms.Integrations.Search.Algolia.Handlers.AlgoliaContentCacheRefresherHandler.RebuildIndex(System.Collections.Generic.IEnumerable`1<Umbraco.Cms.Core.Models.IContent>)
[...]

PR #154, opened 2023-11-28
Implement Algolia converter collection builder.

acoumb:
Current PR contains the refactoring of converters using the existing `CMS` and `Forms` collection builders, allowing developers to append their own custom converters for _Algolia_, or to replace one of the default five converters I've added.

The current default converters are intended for the following Umbraco property editors:
- `Umbraco.TrueFalse`
- `Umbraco.Decimal`
- `Umbraco.Integer`
- `Umbraco.MediaPicker3`
- `Umbraco.Tags`

To create a new converter one should implement the `IConverter` interface, specify the name of the property editor and add the new implementation. Then the new converter will need to be added to the `Algolia Converters` collection.

For example (this will be added to the docs aswell), to remove the existing tags converter, I will ...
1. Create the converter
image

2. Replace existing converter with the new one
image

3. Inject custom converters
image

With this update, I have also addressed these issues:
- #151 
- #150 
- #149 (picked some of the recommendations from this PR #153 )


AndyButland (approved):
Looks good.  A few comments added to consider but once you've looked at those happy to go ahead with this update.

PR #159, opened 2024-02-02
Add Shopify pagination

acoumb:
Current PR contains the implementation of pagination feature for integration with _Shopify_.

The feature is slightly different from a normal pagination feature, due to _Shopify's_ [`relative cursor-based pagination`](https://www.shopify.com/partners/blog/relative-pagination). With version `2019-07` of their API, _Shopify_ no longer accepts a `page` parameter with their requests and have moved to relative pagination.

With relative pagination, I am using the `Link` header from the response to pick the necessary information for moving forward or backwards through the records. Every request that has multiple pages, will include a response header called `Link`. this header contains the URLs for the previous and next pages, if they exist, respecting the sort order asked for. 

The data needed to load the next page is embedded into a query string parameter called `page_info`. The number of records per page is defaulted to 10, using a constant value, and are passed to each request using the `limit` parameter.

Some patterns for the `Link` header response:
- Load first page
Response will contain a `Link` header with details of the next page, in this format:
```
<https://my-umbraco.myshopify.com/admin/api/2024-01/products.json?limit=10&page_info=[*****]>; rel="next"
```
- Load last page
Response will contain a `Link` header with details of the next page, in this format:
```
<https://my-umbraco.myshopify.com/admin/api/2024-01/products.json?limit=10&page_info=[*****]>; rel="previous"
```
- Load a middle page
Response will contain a `Link` header with details of the next page, in this format:
```
<https://my-umbraco.myshopify.com/admin/api/2024-01/products.json?limit=3&page_info=[*****]>; rel="previous", 
<https://my-umbraco.myshopify.com/admin/api/2024-01/products.json?limit=3&page_info=[*****]>; rel="next"
```

Instead of using the actual URLs from the header, I am just picking the `page_info` parameter with specifications of next/previous, and using it in the controller/service.

Other updates included:
- Return products count using a separate HTTP request to `https://my-umbraco.myshopify.com/admin/api/2024-01/products/count.json`. After the configuration is validated, the `vm.pagination.totalPages` property is set with the count.
- When retrieving products for a page, the response object contains two properties (`nextPageInfo` and `previousPageInfo`) with the `page_info` parameter used to toggle pages.
- `umb-pagination` directive used has a different behavior - because I am unable to sent an actual page to the API, we can only move one step ahead or back. This means that if a specific page is selected, I cannot jump to the exact page, I just force the pagination to jump with just one position, simulating the same behavior when pressing `Next`/`Previous`.
- Value converter updated to support new implementation
- `HttpResponseHelper` helper method to parse the adequate `page_info` parameters.

We can have a demo tomorrow in our 1-1, only issue I've encountered is with displaying the selected products in the grid when the overlay is opened (e.g., if I select products from page 3, save them and try to select new products, the list view will show me the first page. Only if I move to the 3rd page again, I will be able to see my selected products).

### DIFF
--- a/src/Umbraco.Cms.Integrations.Commerce.Shopify/Services/IShopifyService.cs
+++ b/src/Umbraco.Cms.Integrations.Commerce.Shopify/Services/IShopifyService.cs
@@ -13,6 +13,10 @@ namespace Umbraco.Cms.Integrations.Commerce.Shopify.Services
 
         void RevokeAccessToken();
 
-        Task<ResponseDto<ProductsListDto>> GetResults();
+        Task<ResponseDto<ProductsListDto>> GetResults(string pageInfo);
+
+        Task<ResponseDto<ProductsListDto>> GetProductsByIds(long[] ids);
+
+        Task<int> GetCount();
     }
 }
--- a/src/Umbraco.Cms.Integrations.Commerce.Shopify/App_Plugins/UmbracoCms.Integrations/Commerce/Shopify/Render/Products.cshtml
+++ b/src/Umbraco.Cms.Integrations.Commerce.Shopify/App_Plugins/UmbracoCms.Integrations/Commerce/Shopify/Render/Products.cshtml
@@ -7,7 +7,10 @@
         {
             <dt class="product-title">@product.Title</dt>
             <dd class="product-body">@Html.Raw(product.Body)</dd>
-            <img class="product-image" src="@product.Image" alt="@product.Title" />
+            @if (!string.IsNullOrEmpty(product.Image))
+            {
+                <img class="product-image" src="@product.Image" alt="@product.Title" />
+            }
         }
     </dl>
 }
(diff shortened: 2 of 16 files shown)


src/Umbraco.Cms.Integrations.Commerce.Shopify/Services/IShopifyService.cs
```diff
@@ -13,6 +13,10 @@ public interface IShopifyService
 
         void RevokeAccessToken();
 
-        Task<ResponseDto<ProductsListDto>> GetResults();
+        Task<ResponseDto<ProductsListDto>> GetResults(string pageInfo);
```
AndyButland, line 16: This will be a breaking change, which is fine for this component, but bump the version to a new minor.

AndyButland (approved):
Code looks good to me.  As you say, give me a quick demo when we catch up tomorrow just to see how it works in use.

acoumb:
With the latest commit, I've used CSS to hide the paginator's numbers. 

Navigation between pages will be handled using `Next/Previous`:
image
image
image

AndyButland (approved):
Looks good to me.

PR #171, opened 2024-02-13
Use correct wording at HubSpot

bjarnef:
The wording of Hubspot should be `HubSpot`.

The classes, properties etc. could eventually be updated, but isn't crucial.

See also: https://github.com/umbraco/Umbraco.Forms.Integrations/pull/97


acoumb (approved):
Looks good, thank you for the updates.

Issue #160, opened 2024-02-02
Dynamics Integration Database Table Issue


RalucaDC:
Hi. 
I have found an issue where the AccessToken column in the dynamicsOAuthConfiguration table is not big enough to store the token.
I am on an older version of Umbraco, 8.18.
Hope this helps someone.
Thanks

---
_This item has been added to our backlog AB#37613_

PR #172, opened 2024-02-13
Bugfix/dynamics access token column size

acoumb:
Current PR adds a migration for increasing the length of the column storing the access token, following issue #160 .

During testing, I was not able to replicate the issue, access tokens received on my account were persisted successfully. I've modified the column to allow a length of 4000, the max value for SQL CE `varchars` (for SQLite not relevant).


AndyButland (approved):
This looks fine.

It doesn't really matter as of course the migration will run, but in Authorized Services we also amended the DTO used to create the table in the original migration step. So perhaps you should also increase the field size [here](https://github.com/umbraco/Umbraco.Cms.Integrations/blob/542f33e4ac83afd51a81dd98421f1a24586260e2/src/Umbraco.Cms.Integrations.Crm.Dynamics/Migrations/DynamicsMigration.cs#L65C17-L65C21), move the 4000 number into a constant, and rename this migration to be called `EnsureAccessTokenColumnLength`?

Up to you though, it'll work fine as is.

PR #175, opened 2024-03-14
Fix Variant VariesByCulture issue

SebastianFalborg:
**Description**
I found an issue, while using the Umbraco Algolia integration, when trying to index content which has nested content.

**How to replicate**
Install a fresh version of Umbraco 13 with the clean starter kit, and add the Umbraco.Cms.Integrations.Search.Algolia package.
To into Umbraco Backoffice and navigate to Algolia Search Management.
Add a new Index Definition, and pick Article.
If you only pick Title and Subtitle, and to back to Build the index, it will work, but if you include Content Rows, you will get this error.
image

**What is fixed in this PR?**
The problem lies in the fact culture and segment is set to be string.Empty, when trying to get the IndexValue for the property, but when validating the property, it only checks if the culture or segment values are null, and not if they are empty strings.

I have changed the two string.Empty values to null, so they get validated correctly later in the code.
More specifically in 
`Umbraco.Cms.Core.PropertyEditors.NestedPropertyIndexValueFactoryBase`
in this part of the Handle method

```
// We want to ensure that the nested properties are set vary by culture if the parent is
// This is because it's perfectly valid to have a nested property type that's set to invariant even if the parent varies.
// For instance in a block list, the list it self can vary, but the elements can be invariant, at the same time.
if (culture is not null)
{
    propertyType.Variations |= ContentVariation.Culture;
}

if (segment is not null)
{
    propertyType.Variations |= ContentVariation.Segment;
}

return propertyType;
```

### DIFF
--- a/src/Umbraco.Cms.Integrations.Search.Algolia/Builders/ContentRecordBuilder.cs
+++ b/src/Umbraco.Cms.Integrations.Search.Algolia/Builders/ContentRecordBuilder.cs
@@ -76,7 +76,7 @@ namespace Umbraco.Cms.Integrations.Search.Algolia.Builders
                     }
                     else
                     {
-                        var indexValue = _algoliaSearchPropertyIndexValueFactory.GetValue(property, string.Empty);
+                        var indexValue = _algoliaSearchPropertyIndexValueFactory.GetValue(property, null);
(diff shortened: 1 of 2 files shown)


acoumb (approved):
Looks good to me. Thank you for addressing this.

PR #181, opened 2024-04-02
Feature: Change Record.Path to be a list of strings

SebastianFalborg:
**Description**
Right now the Path property on Record is a string.
The Path property in the Record class is retrieved from the Path property in the IContent class.
The Path is a comma separated list of ids, and looks something like this "-1,2289,2290,2748".

As it is now. You can't do any filtering by path, because you can't do custom searches like that in Algolia.
By changing the Path property to a List of strings, you can do better searches, by filtering using path.

Examples for this could be
- Hiding content until a user is logged in
- Give the user a dropdown to filter by different path roots
- Hiding content branches entirely using the root path


acoumb (approved):
Looks good to me, thank you.

Issue #180, opened 2024-04-02
ActiveCampaign only returns 20 forms


AndersBrohus:
We face an issue where the ActiveCampaign Form Picker only returns 20 forms, even though there 42 forms created.
Searching does not work since it just filters in the already returned forms.

acoumb:
Hi @AndersBrohus ,

I have released version [2.0.4](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Crm.ActiveCampaign/2.0.4) of the package with support for pagination.

Thank you for bringing this to our attention,
Adrian

AndersBrohus:
Thanks alot @acoumb :)

PR #182, opened 2024-04-03
Add ActiveCampaign Form Picker pagination

acoumb:
Current PR adds pagination to the _ActiveCampaign_ Form Picker, following issue #180 .

By default, the number of results is limited to 20 (with a max of 100), therefor I have introduced a default size of 10 for the number of items displayed on a page, value that I then use as a query string parameter - `limit` - to toggle the number of results. Navigating the forms is managed through an additional query string parameter called `offset`, a zero-based index.

For example, the initial request will be `/api/3/forms?limit=10`, to get the first 10 items, then for other pages the path will be `/api/3/forms?limit=10&offset=[(page - 1)*pageSize]`.

More details on pagination can be found [here](https://developers.activecampaign.com/reference/pagination).

The update can be viewed in action in this [demo](https://us02web.zoom.us/clips/share/AQC5u5TWS0JfPcqRuKNPVZnQVKrolctGOrxDY7j-3i-ZmTUn9_fxJRXAUIXm7B-XMjZz0OVk29DnO73NztfwIiyPBA.e4OVZmwbmDdr0kHh).

### DIFF
--- a/src/Umbraco.Cms.Integrations.Crm.ActiveCampaign.Core/Constants.cs
+++ b/src/Umbraco.Cms.Integrations.Crm.ActiveCampaign.Core/Constants.cs
@@ -9,6 +9,8 @@ namespace Umbraco.Cms.Integrations.Crm.ActiveCampaign.Core
 
         public const string FormsHttpClient = "FormsClient";
 
+        public const int DEFAULT_PAGE_SIZE = 10;
+
         public class Resources
         {
             public const string AuthorizationFailed = "ActiveCampaign authorization failed.";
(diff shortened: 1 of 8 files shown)


src/Umbraco.Cms.Integrations.Crm.ActiveCampaign.Core/Constants.cs
```diff
@@ -9,6 +9,8 @@ public class Constants
 
         public const string FormsHttpClient = "FormsClient";
 
+        public const int DEFAULT_PAGE_SIZE = 10;
```
AndyButland, line 12: Pascal case - `DefaultPageSize` - is the standard for constants in the code base.

AndyButland (approved):
Just a small casing comment.

PR #185, opened 2024-04-12
Bug/add context

gardarthorsteins:
There seems to be an issue running the index builder in background threads like hangfire when you publish nodes with Content Service.

_record.Url = _urlProvider.GetUrl(content.Id) was giving error that the Umbraco Context was not found.

using var ctx = _umbracoContextFactory.EnsureUmbracoContext(); Was set in BuildIndex in SearchController but its missing in AlgoliaContentCacheRefresherHandler.

I changed the code so its passed inside ContentRecordBuilder for it not to be missing again not matter how you call the code.

I reverted the Path change also.


acoumb:
Hi @gardarthorsteins ,

I've manually picked just the changes regarding the Umbraco context and released version [2.1.4](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Search.Algolia/2.1.4) with your update.

Thank you for your input on this.

Adrian

PR #187, opened 2024-04-18
Add extension method for handling lists and arrays.

acoumb:
With this PR, I've tried to address an issue raised from the COMA team about the string representation returned from polling for document types properties (e.g. ```System.Collections.Generic.List`1[Umbraco.Core.Models.PublishedContent.IPublishedContent]``` or `System.String[]`).

My approach to this was to add an extension method for handling lists and arrays using reflection, without having to tightly couple the implementation with types from CMS. This way I could avoid introducing breaking changes and address all versions of the CMS (8+).

I was considering using a similar approach with [`Algolia`](https://github.com/umbraco/Umbraco.Cms.Integrations/blob/main/src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaSearchPropertyIndexValueFactory.cs), but I think, if I remember correctly, that `PropertyEditorCollection` was fully used with `Examine` with version 11 or 12 of the CMS, and am not sure if it will deliver the expected result on Umbraco 8.


AndyButland (commented):
Looks like a smart solution and I'm happy if it solves the issue.  But just wondering if you are assuming a bit much in your reflection code.  I.e. you are expecting a "Name" property to exist by the looks of it, which is fine for `IPublishedContent` but not necessarily going to be there for other types.

So I wonder if you might be better just handling known core types, and for others they aren't supported.  Similar to how you are doing media.  If you support that types that come out of the content picker, multinode picker, nested content etc., that should be OK.

Interested in the use case from COMA here too - they are looking to associate content publish events with something I assume, but would like to here what they are actually looking to solve by this integration.

acoumb:
With the latest commit, I took the content value retrieval out of the `static` context of the extension methods and introduced a service for handling the data. Then, in the service, I am using a content factory to create custom parsers depending on the editor alias.

Issue #188, opened 2024-04-24
Algolia - TinyMCE not indexed in Umbraco 13 with 2.1.4 Algolia


gardarthorsteins:
When indexing TinyMCE properties the value is always return empty to algolia.

This line here is the problem inside of AlgoliaSearchPropertyIndexValueFactory

`var indexValues = propertyEditor.PropertyIndexValueFactory.GetIndexValues(property, culture, null, true);`

The values are in the property but indexValues return empty string.

I'm guessing the issue is the Umbraco update to the TinyMCE property lately, it returns an json object now instead of a string.

I was hoping to fix the problem temporarily by creating a converter for TinyMCE properties but I would need the property data be passed into it to work with the values my self.

I think that passing the property to the converter would be very powerful so we could not only do specific thing with property editors but also target specific ones with the alias or use the raw values or self.

acoumb:
Hi @gardarthorsteins ,

I have included this update with today's [2.1.5](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Search.Algolia/2.1.5) patch release of Algolia.

Thanks,
Adrian

PR #189, opened 2024-04-29
Include IProperty with Algolia's index value converter

acoumb:
Current PR addressed the issue with RTE detailed [here](https://github.com/umbraco/Umbraco.Cms.Integrations/issues/188) for Umbraco 13, by passing `IProperty` to the parse method of the converters. 

This will make it easier for developers to work directly with the content property. And maybe perhaps just send the `IProperty` instead of the `IEnumerable<object> indexValues`?


AndyButland (approved):
Looks OK to me, It is a breaking change, but I don't see a good way to handle it without having to handle a null `IProperty` which would be a bit odd.  So I think if you make it a new minor and add details of the change to the release notes that should be OK.

PR #190, opened 2024-05-16
use correct framework

kasperk81:
the undocumented names (without period in dotnet 5+) can drop their support when move to double-digit (to avoid potential conflicts) https://learn.microsoft.com/en-us/dotnet/standard/frameworks#supported-target-frameworks


AndyButland:
Thanks for the contribution here @kasperk81.

@acoumb - not sure if this is still to merge in, or whether it's been handled in the overlapping PR you've linked above, but please can we either close or merge?  Thanks.

acoumb (approved):
Hi @kasperk81 ,
Thank you for your contribution, looks good to me. I'll merge them in, and then release them once minor/patch versions of the integrations are scheduled.
Regards,
Adrian

Issue #191, opened 2024-05-21
A number of problems on Algolia


thenexus00:
Hi team,
Unfortunately the latest version has a number of issues...

**Backoffice UX:**
- You can not deselect properties. If you deselect an item and hit save and view again it is reselected
- Deselecting a property removes the border but "selected" tag that displays remains.

**Bigger issues with data syncing:**

- You can not send a content block grid. This is too much data and the byte size is way too big.
- As above some other custom property fields run into the error of byte size. Basically as I have seen with some custom projects the raw JSON output contains a lot of the raw Umbraco data that is not needed and thus too much. One example is the recursive parent data contained being there which is too much.

- The other big issue is that it simply is not posting data across when the sync is run in the back office tool:

I have a page type, I select some properties, the API is connected.
While the index is created there are no items.
I can see the request:
Host: c23-eu-2.algolianet.com
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:127.0) Gecko/20100101 Firefox/127.0
Accept: */*
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate, br, zstd
x-algolia-api-key: 14c2****************************
x-algolia-application-id: 3B77DDHVKX
content-type: application/x-www-form-urlencoded
Content-Length: 27
Origin: https://dashboard.algolia.com
Connection: keep-alive
Sec-Fetch-Dest: empty
Sec-Fetch-Mode: cors
Sec-Fetch-Site: cross-site
Sec-GPC: 1
Priority: u=1

The request body though is:
{
  "searchableAttributes": [
  ]
}

Response:
{
  "updatedAt": "2024-05-17T06:21:47.977Z",
  "taskID": 1162837526000
}

**Rolling back to previous versions it works but then there are still other issues...**

You can not properly send media or other property types with meaningful data or values to Algolia.
It relates to the oversize issues with picking some properties as well. There needs to be a curation of all the property types Umbraco offers and work to have curated data going to Algolia and not just a dump.

**Images:**
You mainly need the image path URL and it being Absolute
Since images have crops you should be configuring what crop alias to use or if it is the raw URL

**Files:**
As with images you want an absolute URL file. Basically the media picker you need to know the type and output accordingly be it image or file

**Tags:**
This just does not output in a useable way and the code management when rendering searching and actually getting these to be useful in the search data formation in some cases not possible and others it is just hard.

**Content Grid:**
[...]

PR #192, opened 2024-05-27
Updated convertes, index value retrieval and UI fix.

acoumb:
Current PR addresses some of the issues raised [here](https://github.com/umbraco/Umbraco.Cms.Integrations/issues/191), which impacted the dashboard UI and data indexing.

Two changes are included with this minor release, one to the UI fixing the card deselected event, and the main one to the way converters behave.

I have removed the `indexValues` parameter of `IAlgoliaIndexValueConverter.ParseIndexValues` and work directly with the property. The `IProperty` parameter was included with the previous update of Algolia, and it would make sense to use it directly and not the index values returned by the CMS index value factory. 

To that extend, I moved the converters to run first, then to move to the editor's index value factory from CMS.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Search.Algolia/App_Plugins/UmbracoCms.Integrations/Search/Algolia/js/dashboard.controller.js
+++ b/src/Umbraco.Cms.Integrations.Search.Algolia/App_Plugins/UmbracoCms.Integrations/Search/Algolia/js/dashboard.controller.js
@@ -382,9 +382,9 @@
                     headline: 'Algolia',
                     message: 'Index built successfully'
                 });
-                vm.loading = false;
             }
         });
+        vm.loading = false;
 
         closeBuildDialog();
     }
--- a/src/Umbraco.Cms.Integrations.Search.Algolia/App_Plugins/UmbracoCms.Integrations/Search/Algolia/views/dashboard.html
+++ b/src/Umbraco.Cms.Integrations.Search.Algolia/App_Plugins/UmbracoCms.Integrations/Search/Algolia/views/dashboard.html
@@ -117,10 +117,11 @@
                             <uui-label slot="label">{{ vm.manageIndex.selectedContentType.name }} Properties</uui-label>
                             <uui-icon-registry-essential>
                                 <div class="alg-col-3">
-                                    <uui-card-content-node selectable
+                                    <uui-card-content-node selectable ng-attr-id="{{property.alias}}"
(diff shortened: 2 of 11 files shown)


Brandon-Hale:
Hey Guys, we have run into an issue with the RTE text not being picked up in the algolia data but just an empty string. I believe on Umbraco 13 and version 2.1.5 on the search algolia which in this I can see is moved to 2.2.0 but hasn't been merged as of yet. Trying to find a work around currently but this seems like it would fix issues we have, also cant upgrade to version 3.0.0 as it requires Umbraco 14. Hopefully this can get merged soon :)

acoumb:
Hi @Brandon-Hale ,
Version 2.2.0 of Algolia is still a work in progress, I will revisit it in the following days. In the meantime, maybe it will help with your use case to create a custom converter for the RTE as detailed [here](https://docs.umbraco.com/umbraco-dxp/integrations/algolia#version-2.1.5-and-up).
Thank you,
Adrian

Brandon-Hale:
Thankyou for this, didnt see this before and sure it will be very helpful. Will keep an eye out thanks again

Brandon-Hale:
> Thankyou for this, didnt see this before and sure it will be very helpful. Will keep an eye out thanks again

Just on this the Core Variable in: public string Name => Core.Constants.PropertyEditors.Aliases.Tags;

isnt being recognised any ideas?

acoumb:
You cannot find it in `Umbraco.Cms.Core.Constants.PropertyEditors.Aliases`? You just need the name of the property editor - `Umbraco.Tags`

Brandon-Hale:
Just needed Constants.x.x.xxx think becuase our namespace is Company.Umbraco.xx so it was just messing with the namespace

Brandon-Hale:
`public class AngoliaIntegrationConverter : IAlgoliaIndexValueConverter
{
	public string Name => Constants.PropertyEditors.Aliases.TinyMce;

	public object ParseIndexValues(IProperty property, IEnumerable<object> indexValues)
	{
		return new[] { "Umbraco", "is", "awesome" };
		//return new KeyValuePair<string, object>(property.Alias, indexValues);
	}
}`

Is this all thats needed to supposed to show up in Algolia for an RTE (still getting "")? On version 2.1.5 (https://docs.umbraco.com/umbraco-dxp/integrations/algolia#version-2.1.5-and-up) the issue is mentioned on it but doesnt seem to solve it.

acoumb:
You will need to add your new converter like this
image

AndyButland (approved):

PR #193, opened 2024-06-07
Feature/algolia v14

acoumb:
Current PR contains the updates for Umbraco 14 for `Algolia` integration, with some bugfixes and design updates.

This is also to get feedback on the release-ready status of the integration based on your experience with the other products.


src/Umbraco.Cms.Integrations.Search.Algolia/Api/Management/Controllers/BuildIndexController.cs
```diff
@@ -0,0 +1,80 @@
...
+                            .BuildFromContent(contentItem, (p) => contentDataItem.Properties.Any(q => q.Alias == p.Alias))
+                            .Build();
+
+                        payload.Add(record);
+                    }
+                }
+
+                var result = await IndexService.PushData(index.Name, payload);
+
+                return new JsonResult(result);
```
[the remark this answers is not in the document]
acoumb: Updated

src/Umbraco.Cms.Integrations.Search.Algolia/Client/public/umbraco-package.json
```diff
@@ -0,0 +1,12 @@
+﻿{
+  "name": "Umbraco.Cms.Integrations.Search.Algolia",
```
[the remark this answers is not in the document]
acoumb: Updated

src/Umbraco.Cms.Integrations.Search.Algolia/Client/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "algolia",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc && vite build",
+    "preview": "vite preview"
```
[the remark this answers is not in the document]
acoumb: Updated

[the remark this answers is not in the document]
AndyButland: I don't see the generated files though?  You will still need to check them in so they are there when the package is built/is installed.

src/Umbraco.Cms.Integrations.Search.Algolia/Api/Management/Controllers/SaveIndexController.cs
```diff
@@ -56,11 +58,12 @@ public async Task<IActionResult> SaveIndex([FromBody] IndexConfiguration index)
                     ? Result.Ok()
                     : await IndexService.PushData(index.Name);
 
-                return new JsonResult(result);
+                return Ok(result);
             }
             catch (Exception ex)
             {
-                return new JsonResult(Result.Fail(ex.Message));
+                Logger.LogError(ex, ex.Message);
```
AndyButland, line 65: Wondering if you really need the `try/catch`?  As I think any Umbraco exception will get logged anyway.

src/package.json
```diff
@@ -9,6 +9,10 @@
   "author": "",
   "license": "ISC",
   "devDependencies": {
+    "@umbraco-cms/backoffice": "^14.0.0--preview007",
```
AndyButland, line 12: This is still referencing a preview version.

AndyButland: I guess it and the package-lock.json next to it isn't needed right?  As the one you are using is in "Client".

src/Umbraco.Cms.Integrations.Search.Algolia/Api/Management/Controllers/SaveIndexController.cs
acoumb: Confirmed, Umbraco will log the exceptions.

src/Umbraco.Cms.Integrations.Search.Algolia/App_Plugins/UmbracoCms.Integrations/Search/Algolia/config/lang/en-US.xml
AndyButland, line 1: Noticed this was deleted, but don't see an equivalent.  Maybe you are just leaving it hardcoded in English, which is fine, or maybe you could look at adding localization in the new way (see Forms for an example).

src/Umbraco.Cms.Integrations.Search.Algolia/Client/package.json
```diff
@@ -0,0 +1,27 @@
...
+  },
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc && vite build",
+    "preview": "vite preview",
+    "watch": "vite build --watch",
+    "generate:api": "openapi-ts --input https://localhost:44340/umbraco/swagger/AlgoliaSearch/swagger.json --output src/generated --debug true --enums typescript --lint true --schemas false"
+  },
+  "dependencies": {
+    "lit": "^3.1.2"
```
AndyButland, line 18: Not sure if we need this?  See [the Forms version of this file](https://github.com/umbraco/Forms/blob/v14/dev/src/Umbraco.Forms.StaticAssets/package.json) for comparison.

src/Umbraco.Cms.Integrations.Search.Algolia/Client/src/api/core/ApiError.ts
AndyButland, line 1: The files under `/api/` are all generated now right?  Might be good to just delete anything you had there before, and re-generate, just to make sure the generation is creating everything we need.

For consistency with the products maybe put it under `src/generated` too?

src/Umbraco.Cms.Integrations.Search.Algolia/Client/src/models/AlgoliaContentTypeModel.ts
AndyButland, line 1: Could/should these not be generated too?

AndyButland (commented):
Looking good to me... but still a few comments just as I'm not sure what's being generated here.  With Forms we've been able to generate the front-end representation of the types of the API's request and response models, as well as the services that call the API.  So would just like to align that with what we have here.

Might need to be next week, but would like @nathanwoulfe to have a look over too.

src/Umbraco.Cms.Integrations.Search.Algolia/Client/package.json
nathanwoulfe: Yup, we don't need to take a direct dependency on Lit - the backoffice package provides it as an export 👍

acoumb: Removed dependency.

acoumb:
Fixed api resources generation by using the Swagger JSON schema on `http` instead of `https`, and updated all references.

nathanwoulfe:
Looks like I've made a stack of changes, but most of it is just linting/style to align with the other products.
 - I've updated the API controllers to remove the protected properties on the base controller and instead have each controller provide its own dependencies - slims down the controllers as they only deal with direct concerns.
 - found my stack overflow issue from yesterday - `GetContentTypesController` had an endpoint method named `GetContentTypes` and a private method with the same name accepting an optional int ID argument. Former attempted to call the later, but fell into a self-referencing loop. Renamed a couple of methods and we're all good.
 - added the eslint rules used in other projects, which I originally stole from CMS, again just for consistency
 - moved generated files from /src/api to /generated, again just for consistency
 
Couple of remaining things to consider:
 - do we need a context at all? The dashboard element could directly new up a repository and away we go. Context isn't storing any stateful data, but I suppose it's going to be more performant to not init the repo and data-source on every route event.
 - given the above, if we keep the context we could remove the data-source and have the repo call the services directly. Repo-datasource-store pattern is useful, but we don't have to use it if we don't get any benefit from it
 - would be good to look at moving the error handling down to the repository layer (rather than the dashboard/views) as we can then manage error notifications in the repo, and provide only data back to the views, which would simplify these quite a bit, eg:
 ```ts
await this.#algoliaIndexContext?.getIndices()
            .then(response => {
                this._indices = response as Array<IndexConfigurationModel>;
                this._loading = false;
            })
            .catch(error => this.#showError(error.message));
```
can use async/await without the callbacks:
```ts
this._indices = await this.#algoliaIndexContext?.getIndices();
this._loading = false;
```

Long story short, this is looking great and ready to go (things to consider above can be done at any time, if we want to)

acoumb:
@nathanwoulfe I've removed the data source and moved the logic and error handling to the repository scope. Also I made some small fixes along the way. Could you have a quick look over things now before merging them? 
Thank you.

nathanwoulfe:
@acoumb looks good to me - I made a couple more changes
 - use type for context rather than the element reference (means one fewer import)
 - de-structure the response from the repository (means one fewer variable each call)
 - mainly because I wanted to see if it was possible, the selected content types and properties are now highlighted on load.
 - use `umb-property-layout` rather than the form layout, it is styled better for the backoffice

nathanwoulfe (approved):

PR #194, opened 2024-06-11
Added support for displaying prices and filtering on products picked

RachBreeze:
Hello

We have a requirement to display the price of products which have been picked from shopify. This PR is to help resolve this.

I also think the fact that product type and tags are missing from the data returned by the integration means it's not possible in Umbraco to create searches that use the tags. I did consider mapping tags to a `ienumerable<string>` and am happy to do this.

Finally I have added Status in as the integration could result in a mapping to products that are no longer published. I did consider mapping `status` to https://shopify.dev/docs/api/admin-graphql/2024-04/enums/ProductStatus and can do, but also it would mean the API would need further work if future status' were added.


acoumb (approved):
Looks good to me, thank you @RachBreeze

Issue #197, opened 2024-06-21
Null error when trying to retrieve Dynamics365 forms


ben-umble:
Hi,

I'm trying to use the Dynamics365 integration to display forms on my website. I've installed version 1.3.3 and authenticated okay but when I try to add a form I get a null instance error.

# Useful Details

1. Umbraco version: `9.5.4`
2. Umbraco.Cms.Integrations.Crm.Dynamics version: `1.3.3`
3. I've tried all options for the `Modules` setting in the data type with the same result
4. I've tried downgrading to version 1.3.2 and I do not get the error (no forms show but I believe that's more to do with our Dynamics365 dev instance than a plugin problem).

# Error

```
Received an error from the server
An error occurred
Object reference not set to an instance of an object.

Exception Details
System.NullReferenceException, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e: Object reference not set to an instance of an object.
Stacktrace
at Umbraco.Cms.Integrations.Crm.Dynamics.Services.DynamicsService.GetForms(DynamicsModule module)
   at Umbraco.Cms.Integrations.Crm.Dynamics.Controllers.FormsController.GetForms(String module)
   at lambda_method2510(Closure , Object )
   at Microsoft.AspNetCore.Mvc.Infrastructure.ActionMethodExecutor.AwaitableObjectResultExecutor.Execute(IActionResultTypeMapper mapper, ObjectMethodExecutor executor, Object controller, Object[] arguments)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeActionMethodAsync>g__Awaited|12_0(ControllerActionInvoker invoker, ValueTask`1 actionResultValueTask)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeNextActionFilterAsync>g__Awaited|10_0(ControllerActionInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.Rethrow(ActionExecutedContextSealed context)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.Next(State& next, Scope& scope, Object& state, Boolean& isCompleted)
[...]

PR #198, opened 2024-06-26
Fix NULL reference

acoumb:
Current PR fixes issue #197 by handling NULL reference when handling the list of Dynamics forms.


AndyButland (commented):
Fix looks fine, but seems you should at least be logging an error if you have a null value here.  Looks like this happens when you don't get a success code response from the API call, but in that case, it would be good to log what the problem is.

Kind of like we do [here](https://github.com/umbraco/Umbraco.AuthorizedServices/blob/main/src/Umbraco.AuthorizedServices/Services/Implement/AuthorizedServiceCaller.cs#L114).  Not saying you should refactor the attempt pattern but via this link just showing the type of information that would be useful to log.

acoumb:
Added logging to the HTTP method to persist the error message before returning NULL.

AndyButland (approved):
Looks good, thanks.

PR #199, opened 2024-07-03
Feature/hubspot v14

acoumb:
Current PR contains the redesigned HubSpot Forms integration for Umbraco 14. It follows the same flow as the other version of the integration, but for the following features that I could not find a resolution.
1. Rendering the authorization component in the modal view, to allow editors to authenticate from the content item, without having to go into the settings. The element is referenced as <hubspot-authorization></hubspot-authorization>, but does not render on most of the occasions.
Instead of this: 
image
I get:
image

2. Localization of the `Add` button is not working sometimes. 
image

3. Unable to render the picked HubSpot form, although the property of the model is the correct one:

image


nathanwoulfe (approved):
Hey @acoumb, I think my changes from earlier should resolve the issues you mentioned - had to export the auth element to ensure the JS file exists, so that the browser can add the element to the custom element registry.

I also let VS Code do its thing with linting, and applied pretty much everything it suggested.

src/Umbraco.Cms.Integrations.Crm.Hubspot/Client/src/property-editor/authorization-property-editor.element.ts
```diff
@@ -0,0 +1,189 @@
...
+        });
+    }
+
+    async connectedCallback() {
+        super.connectedCallback();
+        await this.#checkApiConfiguration();
+    }
+
+    async #checkApiConfiguration() {
+        const { data } = await this.#hubspotFormsContext.checkApiConfiguration();
```
nathanwoulfe, line 61: Could consider moving this back to the context, and making the serviceStatus object an observable. Then the dependent components can observe that object rather than having to request the API configuration

AndyButland:
Couple of comments from me.  I wasn't able to fully complete testing it (might be user error).

1. Not quite sure what the flow is intended to be for local development, but I found I had to copy `\src\Umbraco.Cms.Integrations.Crm.Hubspot\wwwroot` into `\src\Umbraco.Cms.Integrations.Testsite.V14` to get the front-end to show up.  Is there a good way to automate that somehow when you do a build?

2. The test site has no reference to the HubSpot project.  Was easy to add, but might be good that this is there by default when pulling down the repository.

3. The API request to check configuration seems OK, but I get this message in the backoffice data type:

```
No API or OAuth configuration could be found. Please review your settings before continuing.
```

Have this in configuration:

```
"Umbraco": {
  "CMS": {
    "Integrations": {
      "Crm": {
        "Hubspot": {
          "OAuthSettings": {
            "ClientId": "...",
            "ClientSecret": "..."
          }
        }
      }
    }
  }
},
```

src/Umbraco.Cms.Integrations.Crm.Hubspot/Client/src/property-editor/authorization-property-editor.element.ts
acoumb: Moved the object into the context and `observe` it from the other components.

acoumb:
Hi @nathanwoulfe ,
I've added some additional changes, if you wouldn't mind checking them out before release:
- as you've suggested, I've moved the settings model into the context as observable, and pick it in the other components without calling the context method.
- dispatch custom events for connect/revoke, so in the modal component of the property editor the forms table would update accordingly.
- removed all references to `Newtonsoft.Json` to align with the overall strategy.

AndyButland:
Few more thoughts on the PR:

- I don't seem to be able to get OAuth working, even when I'm already logged into HubSpot.  The dialog closes but I don't see any update to the "No API or OAuth configuration could be found. Please review your settings before continuing." message.
- Should we align this with Algolia and Shopify for 14 to combine to a single project?  As we likely don't need this now we have an RCL (see discussion [here](https://umbraco.slack.com/archives/C07AP3QAAAW/p1720680939355439)).

acoumb:
With latest commits I've fixed the OAuth issue (a small typo in the authorization component) and fixed an issue with the build pipeline. Additional changes:
- merged the Core project into the RCL
- added option to manually enter the authorization code to be exchanged for access token when the child authorization window has not closed for 7 seconds.

One thing that I could not get to pack is the `/App_Plugins/HubspotForms/Render/HubspotForms.cshtml` razor view. On `dotnet pack` the resource is not included.

acoumb:
Updated the property editor to apply the form details on modal form select.

A demo of the integration: https://us02web.zoom.us/clips/share/9vxJISckoSDwiOVP2I4kRDF8HtUNYiztmXLFxYMIh3OO72WfUSFI2l_ykTdE-WuWY1-9wPZ-vYPr0V83JEUAtTWgcA.lhH6kHDHpgMTVE-s

AndyButland (approved):

PR #200, opened 2024-07-05
V14 integrations: Shopify

dtng95:
- This PR is gonna be a pretty big PR so I'm gonna push my changes everyday so that you guys can review my changes daily.


src/Umbraco.Cms.Integrations.Commerce.Shopify.Core.New/Api/Management/Controllers/CheckConfigurationController.cs
```diff
@@ -0,0 +1,33 @@
...
+    [ApiVersion("1.0")]
+    [ApiExplorerSettings(GroupName = Constants.ManagementApi.GroupName)]
+    public class CheckConfigurationController : ShopifyControllerBase
+    {
+        public CheckConfigurationController(IOptions<ShopifySettings> shopifySettings, IShopifyService shopifyService, IShopifyAuthorizationService shopifyAuthorizationService) : base(shopifySettings, shopifyService, shopifyAuthorizationService)
+        {
+        }
+
+        [HttpGet("check-configuration")]
+        [ProducesResponseType(StatusCodes.Status200OK)]
```
AndyButland, line 26: This should be indicating the type that is returned (i.e. we aren't just returning a 200 response with an empty body here, rather a 200 with some response content too).

src/Umbraco.Cms.Integrations.Commerce.Shopify.Core.New/Api/Management/Controllers/GetAuthorizationUrlController.cs
```diff
@@ -0,0 +1,34 @@
...
+    [ApiVersion("1.0")]
+    [ApiExplorerSettings(GroupName = Constants.ManagementApi.GroupName)]
+    public class GetAuthorizationUrlController : ShopifyControllerBase
+    {
+        public GetAuthorizationUrlController(IOptions<ShopifySettings> shopifySettings, IShopifyService shopifyService, IShopifyAuthorizationService shopifyAuthorizationService) : base(shopifySettings, shopifyService, shopifyAuthorizationService)
+        {
+        }
+
+        [HttpGet("get-authorization-url")]
+        [ProducesResponseType(StatusCodes.Status200OK)]
```
AndyButland, line 27:
```suggestion
        [ProducesResponseType(typeof(string), StatusCodes.Status200OK)]
```

src/Umbraco.Cms.Integrations.Commerce.Shopify.Core.New/Api/Management/Controllers/GetListByIdsController.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    [ApiVersion("1.0")]
+    [ApiExplorerSettings(GroupName = Constants.ManagementApi.GroupName)]
+    public class GetListByIdsController : ShopifyControllerBase
+    {
+        public GetListByIdsController(IOptions<ShopifySettings> shopifySettings, IShopifyService shopifyService, IShopifyAuthorizationService shopifyAuthorizationService) : base(shopifySettings, shopifyService, shopifyAuthorizationService)
+        {
+        }
+
+        [HttpPost("get-list-by-ids")]
```
AndyButland, line 25: Feels like this should be a GET rather than a POST, given it's returning read-only results and not making any updates.

src/Umbraco.Cms.Integrations.Commerce.Shopify.Core.New/Services/UmbracoAuthorizationService.cs
```diff
@@ -27,7 +27,7 @@ public class UmbracoAuthorizationService : BaseAuthorizationService, IShopifyAut
 
         public const string ServiceAddressReplace = "service_address_shop-replace";
 
-        public const string OAuthProxyBaseUrl = "https://hubspot-forms-auth.umbraco.com/";  // for local testing: "https://localhost:44364/";
+        public const string OAuthProxyBaseUrl = "https://localhost:44340/";  // for local testing: "https://localhost:44364/";
```
AndyButland, line 30: Make sure to revert this before merging.

src/Umbraco.Cms.Integrations.Commerce.Shopify.Core.New/ShopifyComposer.cs
```diff
@@ -0,0 +1,45 @@
...
+            // Generate Swagger documentation for Shopify API
+            builder.Services.Configure<SwaggerGenOptions>(options =>
+            {
+                options.SwaggerDoc(
+                    Constants.ManagementApi.ApiName,
+                    new OpenApiInfo
+                    {
+                        Title = Constants.ManagementApi.ApiTitle,
+                        Version = "Latest",
+                        Description = $"Describes the {Constants.ManagementApi.ApiTitle} available for handling indices."
```
AndyButland, line 38: This description is copied from the Algolia integration and should be updated to describe the Shopify package.

src/Umbraco.Cms.Integrations.Commerce.Shopify/App_Plugins/UmbracoCms.Integrations/Commerce/Shopify/Render/ProductsV9.cshtml
```diff
@@ -1,4 +1,4 @@
-﻿@inherits Umbraco.Cms.Web.Common.Views.UmbracoViewPage<List<Umbraco.Cms.Integrations.Commerce.Shopify.Models.ViewModels.ProductViewModel>>
+﻿@* @inherits Umbraco.Cms.Web.Common.Views.UmbracoViewPage<List<Umbraco.Cms.Integrations.Commerce.Shopify.Models.ViewModels.ProductViewModel>>
```
AndyButland, line 1: We remove this presumably since we won't be multi-targeting between different Umbraco major versions with this release.

src/Umbraco.Cms.Integrations.Commerce.Shopify/Client/src/my-element.ts
```diff
@@ -0,0 +1,127 @@
+import { LitElement, css, html } from 'lit'
```
AndyButland, line 1: Should remove (either now, or later if it's useful to refer to for the moment).

src/Umbraco.Cms.Integrations.Commerce.Shopify/Pages/Error.cshtml
```diff
@@ -0,0 +1,26 @@
+﻿@page
```
AndyButland, line 1: Not sure why these Razor pages have been added?

src/Umbraco.Cms.Integrations.Commerce.Shopify/Umbraco.Cms.Integrations.Commerce.Shopify.csproj
```diff
@@ -1,80 +1,81 @@
...
+    <TargetFramework>net8.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <OutputType>Library</OutputType>
+  </PropertyGroup>
 
-	<ItemGroup Condition="'$(TargetFramework)' == 'net472'">
-		<PackageReference Include="UmbracoCms.Web" version="8.5.4" />
-	</ItemGroup>
+  <ItemGroup>
```
AndyButland, line 10: Looks like this needs a clean-up.

AndyButland (commented):
Just had a look over @dtng95 - the core code you've added for the management API and front-end implementation looks good to me.  But there's quite a lot of "extra stuff" in this PR that I'm not quite sure where it's come from - I assume it's from you creating a new project to use, but not an empty one.  

When you've had chance to clean-up - ideally returning to the original folder structure and project names - please let me know you are happy where you have got to. I'll pull it down and have a test.

src/Umbraco.Cms.Integrations.Commerce.Shopify/wwwroot/App_Plugins/Shopify/amount-property-editor.element-eAoDmYox.js
```diff
@@ -0,0 +1,34 @@
+import { UmbElementMixin as u } from "@umbraco-cms/backoffice/element-api";
```
dtng95, line 1: I believe we don't need this file (and others in this folder).

src/Umbraco.Cms.Integrations.Commerce.Shopify/Client/src/property-editor/manifests.ts
```diff
@@ -0,0 +1,47 @@
...
+					label: 'Authorization',
+					description: 'Authorize your Shopify connection.',
+					propertyEditorUiAlias: 'Shopify.PropertyEditorUi.Authorization',
+                },
+                {
+                    alias: 'amount',
+					label: 'Amount',
+					description: 'Set a required range of items selected.',
+					propertyEditorUiAlias: 'Shopify.PropertyEditorUi.Amount',
+                    config: [
```
dtng95, line 28: This is the config for the required range of amount for the selected product list. The min and max values here are successfully displayed in the setting of the Shopify Product Picker data type, but I can't pass these to the property editor.

src/Umbraco.Cms.Integrations.Commerce.Shopify/Client/src/modal/shopify-products-modal.element.ts
```diff
@@ -0,0 +1,284 @@
...
+                    columnAlias: "sku",
+                    value: product.variants.map(v => v.sku).join(","),
+                },
+                {
+                    columnAlias: "barcode",
+                    value: product.variants.map(v => v.barcode).join(","),
+                },
+                {
+                    columnAlias: "price",
+                    value: product.variants[0].price,
```
dtng95, line 231: Is there any product that contains 2 or more variants?

src/Umbraco.Cms.Integrations.Commerce.Shopify/Client/src/modal/shopify-products-modal.element.ts
```diff
@@ -0,0 +1,284 @@
...
+                },
+            ]
+            }
+		});
+	}
+
+    _onSubmit() {
+        if(this._numberOfSelection > 2){
+            this._showError("Cannot select more than 2 products."!);
```
dtng95, line 240: Hardcoded to demonstrate the feature of selecting a limited number of products

src/Umbraco.Cms.Integrations.Commerce.Shopify/Client/src/modal/shopify-products-modal.element.ts
```diff
@@ -0,0 +1,284 @@
...
+                        .items=${this._tableItems}
+                        .selection=${this._selection}
+                        @selected="${this.#onSelected}"
+                        @deselected="${this.#onDeselected}"></umb-table>
+                    
+                </uui-box>
+
+                <div class="maximum-selection">
+                    <span>
+                        Add up to 2 items(s)
```
dtng95, line 268: Also hardcoded to demonstrate the feature of selecting a limited number of products

src/Umbraco.Cms.Integrations.Commerce.Shopify/Client/src/property-editor/manifests.ts
dtng95: Another approach I can think of is to create 2 new settings for Shopify Product Picker: Min and Max values. With that, I won't have to create a brand new property editor like current approach, just 2 simple `Umb.PropertyEditorUi.Integer`. It would be much simpler (I think!)

acoumb:
Hi @dtng95 ,

Tested the _Shopify_ integration and thought about the following points that would be worth revisiting:

- the product picker component should only display the selected products, the _Add_ button or a notification message about the authorization status (in case the Shopify environment is not accessible). This is the default behavior for other _pickers_ in Umbraco, so maybe we can stick to the same. Shopify currently supports access token/outh based access control, but it has been used mainly with access tokens, because the Umbraco Shopify app has slim chances to get published on their marketplace. Therefor an _OAuth2_ authorization component could be included with a future release if we make progress with the app submission.
image
- could you include a _Remove_ button here, so users can edit the list of products without opening again the modal?
image
- these Angular resources can be removed and the rendering view can be used in a similar way with [HubSpot](https://github.com/umbraco/Umbraco.Cms.Integrations/tree/feature/hubspot-v14/src/Umbraco.Cms.Integrations.Crm.Hubspot/Views/HubspotForms/Render)
image
- property editor value validation could be moved a little bit down. If no product is selected, the _Add_ button is not visible. Also this way the user can get a status of the API configuration
image
- a very good [suggestion](https://github.com/umbraco/Umbraco.Cms.Integrations/blob/a8ec3ca2b2bed3a76845406ccd8b176a9babfe59/src/Umbraco.Cms.Integrations.Crm.Hubspot/Client/src/context/hubspot-forms.context.ts#L31) I received from @nathanwoulfe was moving the `checkApiConfiguration` into an observable property of the context and use that property in the components, without having to make multiple calls to the server.

### Management API
image
- I think you can remove the `get` keyword from some of the routes, for example `get-access-token` uses a `POST` request and can be confusing. As a convention, `get-access-token` = `access-token`, `get-list` = `list` and so on.
image
- `GetAuthorizationUrlController` should return an `IActionResult` for consistency with the other endpoints
- `GetListController` - could we use the route as `/list/{pageInfo}` ?
- `RefreshAccessTokenController` - could the route be `refresh-token` to be a bit explanatory? You could perhaps as a general rule use a convention for endpoints that reflect access token actions, `*-access-token` or `*-token`?

dtng95:
> Hi @dtng95 ,
> 
> Tested the _Shopify_ integration and thought about the following points that would be worth revisiting:
> 
> * the product picker component should only display the selected products, the _Add_ button or a notification message about the authorization status (in case the Shopify environment is not accessible). This is the default behavior for other _pickers_ in Umbraco, so maybe we can stick to the same. Shopify currently supports access token/outh based access control, but it has been used mainly with access tokens, because the Umbraco Shopify app has slim chances to get published on their marketplace. Therefor an _OAuth2_ authorization component could be included with a future release if we make progress with the app submission.
>   image
> * could you include a _Remove_ button here, so users can edit the list of products without opening again the modal?
>   image
> * these Angular resources can be removed and the rendering view can be used in a similar way with [HubSpot](https://github.com/umbraco/Umbraco.Cms.Integrations/tree/feature/hubspot-v14/src/Umbraco.Cms.Integrations.Crm.Hubspot/Views/HubspotForms/Render)
>   image
> * property editor value validation could be moved a little bit down. If no product is selected, the _Add_ button is not visible. Also this way the user can get a status of the API configuration
>   image
> * a very good [suggestion](https://github.com/umbraco/Umbraco.Cms.Integrations/blob/a8ec3ca2b2bed3a76845406ccd8b176a9babfe59/src/Umbraco.Cms.Integrations.Crm.Hubspot/Client/src/context/hubspot-forms.context.ts#L31) I received from @nathanwoulfe was moving the `checkApiConfiguration` into an observable property of the context and use that property in the components, without having to make multiple calls to the server.
> 
> ### Management API
> image
> 
> * I think you can remove the `get` keyword from some of the routes, for example `get-access-token` uses a `POST` request and can be confusing. As a convention, `get-access-token` = `access-token`, `get-list` = `list` and so on.
>   image
> * `GetAuthorizationUrlController` should return an `IActionResult` for consistency with the other endpoints
> * `GetListController` - could we use the route as `/list/{pageInfo}` ?
> * `RefreshAccessTokenController` - could the route be `refresh-token` to be a bit explanatory? You could perhaps as a general rule use a convention for endpoints that reflect access token actions, `*-access-token` or `*-token`?

Hi @acoumb ,
I've updated most of your comments, except for the `/list/{pageInfo}`. From my perspective, we don't usually pass any parameter to `pageInfo`, hence it will usually be null or empty string, thus moving the param to query string wouldn't be reasonable.

acoumb:
Hi @dtng95 ,

I've added a couple of updates, feel free to test them out and let me know if you encounter issues.
- Added pagination to the list of products. Currently, the pagination control does not support hiding the buttons, which would have been helpful in the case of Shopify (you can only move one step forward/backwards when going through the list of products), as First/Last buttons are not needed. Therefor, whatever pagination action you select, it will move you one step.
- Cleaned up namespaces
- Updated the persistence of selected products - instead of saving the entire array of products, I will persist only the ids. Then use them to pull the details of the selected products.
- Updates to rendering component and converter based on the previous point.

The only thing remaining I would say it's the removal of `Newtonsoft.Json` and use `System.Text.Json` to align with the other products.

If you've started _Zapier_, let me know and I can take care of the removal.

Thanks

acoumb:
Some final changes to the v14 integration: 
- update package version to `2.0.0`
- cleaned up some unused references and removed `Newtonsoft.Json`
- update view models to align with PRs #194 and #201

PR #202, opened 2024-07-25
42318 V14 Integration: Zapier

dtng95:
- Like the PR for Shopify integration, I will update this PR daily
- The first commit was to create management API for Zapier, and also the structure of the project.


src/Umbraco.Cms.Integrations.Automation.Zapier/wwwroot/css/site.css
```diff
@@ -0,0 +1,22 @@
+html {
```
dtng95, line 1: all files in this wwwroot folder will be removed in the next commit.

src/Umbraco.Cms.Integrations.Automation.Zapier/Api/Management/Controllers/PollingController.cs
```diff
@@ -0,0 +1,68 @@
...
+        public IActionResult GetContentByType(string alias)
+        {
+            //if (!IsAccessValid()) 
+            //    return NotFound();
+
+            //var contentType = _contentTypeService.Get(alias);
+            //if (contentType == null) 
+            //    return Ok(new List<Dictionary<string, string>>());
+
+            //var contentItems = _umbracoHelper.ContentAtXPath("//" + alias)
```
dtng95, line 55: I comment this endpoint out temporarily because `ContentAtXPath` method has been removed. I will have to ask CMS Core team about this.
This is the thread that I found about this removal.
https://github.com/umbraco/Announcements/issues/10

dtng95:
@acoumb Do we need to update the appsettings file in V14 test site, in order to have the settings for Zapier, like we did for Algolia and Shopify (only the settings, no values).

acoumb:
Sure, you can do that.

acoumb:
Hi @dtng95 ,

Implementation looks good, I made a couple of updates and cleaned up a little bit as detailed below:
* broke down controllers into specific API endpoints to align with the other products. As a result, I've prepared a new major version of the Zapier Umbraco app to match the new endpoints.
* updated client API resources
* removed this line of code `options.OperationFilter<BackOfficeSecurityRequirementsOperationFilter>();` as it generated an `ArgumentException: An item with the same key has already been added. Key: 401` on my side. Could not figure out why.
* cleaned up components UI.

Could you have a try at it with the new updates and see if it works alright for you as well? (use version 3.0.0 of the Zapier Umbraco app).

Thanks for the great work,
Adrian

PR #205, opened 2024-08-06
V14: Integrations (Dynamics)

dtng95:
- Add new project for Dynamics Integration
- Introduce new management API
- Add Client folder for front end implementation


src/Umbraco.Cms.Integrations.Commerce.Shopify/Client/config.outputPath.js
```diff
@@ -1 +1 @@
-export const outputPath = 'Release' !== 'Release' ? '../wwwroot' : '../obj/Release/net8.0/clientassets'
```
dtng95, line 1: Do we need to keep this file or it can be removed?

src/Umbraco.Cms.Integrations.Crm.Dynamics/Api/Management/Controllers/CheckOAuthConfigurationController.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    public class CheckOAuthConfigurationController : FormsControllerBase
+    {
+        public CheckOAuthConfigurationController(IOptions<DynamicsSettings> options, DynamicsService dynamicsService, DynamicsConfigurationService dynamicsConfigurationService, DynamicsComposer.AuthorizationImplementationFactory authorizationImplementationFactory) : base(options, dynamicsService, dynamicsConfigurationService, authorizationImplementationFactory)
+        {
+        }
+
+        [HttpGet("check-oauth-configuration")]
+        [ProducesResponseType(typeof(OAuthConfigurationDto), StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status404NotFound)]
```
AndyButland, line 23: Doesn't look like this endpoint can return 404, so think this should be removed.

I won't repeat this comment in other cases, but would be worth just looking carefully over your endpoints and make sure they all have the correct settings for the `ProducesResponseType` attributes.

src/Umbraco.Cms.Integrations.Crm.Dynamics/Api/Management/Controllers/FormsControllerBase.cs
```diff
@@ -0,0 +1,52 @@
...
+    [ApiVersion("1.0")]
+    [ApiExplorerSettings(GroupName = Constants.ManagementApi.GroupName)]
+    [Route($"{Constants.ManagementApi.RootPath}/v{{version:apiVersion}}/forms")]
+    public class FormsControllerBase : DynamicsControllerBase
+    {
+        // Using a static HttpClient (see: https://www.aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/).
+        //private readonly static HttpClient s_client = new HttpClient();
+
+        // Access to the client within the class is via ClientFactory(), allowing us to mock the responses in tests.
+        //public static Func<HttpClient> ClientFactory = () => s_client;
```
AndyButland, line 27: Seems like these comments are no longer relevant and can be removed.

src/Umbraco.Cms.Integrations.Crm.Dynamics/Api/Management/Controllers/FormsControllerBase.cs
```diff
@@ -0,0 +1,52 @@
...
+        protected readonly DynamicsSettings DynamicsSettings;
+
+        protected readonly IDynamicsAuthorizationService AuthorizationService;
+
+        protected readonly DynamicsService DynamicsService;
+
+        protected readonly DynamicsConfigurationService DynamicsConfigurationService;
+
+        public FormsControllerBase(IOptions<DynamicsSettings> options,
+            DynamicsService dynamicsService,
```
AndyButland, line 38: Might be one for @acoumb when he is back.  But seems we might be more idiomatic by defining an interface has and injecting that, rather than the concrete type.  Same for the other services provided as dependencies here.

src/Umbraco.Cms.Integrations.Crm.Dynamics/Api/Management/Controllers/GetAccessTokenController.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Umbraco.Cms.Integrations.Crm.Dynamics.Api.Management.Controllers
+{
+    public class GetAccessTokenController : FormsControllerBase
+    {
+        public GetAccessTokenController(IOptions<DynamicsSettings> options, DynamicsService dynamicsService, DynamicsConfigurationService dynamicsConfigurationService, DynamicsComposer.AuthorizationImplementationFactory authorizationImplementationFactory) : base(options, dynamicsService, dynamicsConfigurationService, authorizationImplementationFactory)
+        {
+        }
+
+        [HttpPost("get-access-token")]
```
AndyButland, line 21: I recall Adrian made some suggestions and changes to align the routes defined on the Shopify management API - and for this one suggestion was to remove the `get-` prefix here (so it became [this](https://github.com/umbraco/Umbraco.Cms.Integrations/blob/d43cffdd74fd6fe904fcf928b11e009f62001ed3/src/Umbraco.Cms.Integrations.Commerce.Shopify/Api/Management/Controllers/GetAccessTokenController.cs#L19)).  So can you do that please, and also compare to the other integrations we have to try to make sure all are as similar as they can be?

src/Umbraco.Cms.Integrations.Crm.Dynamics/Client/src/my-element.ts
AndyButland, line 1: Looks like test code - should remove.

src/Umbraco.Cms.Integrations.Crm.Dynamics/DynamicsFormPickerPropertyEditor.cs
AndyButland, line 1: Don't forget you will still need an adapted version of this class, to define the server-side aspects of your property editor.  See [this file](https://github.com/umbraco/Umbraco.Cms.Integrations/blob/main-v14/src/Umbraco.Cms.Integrations.Commerce.Shopify/Editors/ShopifyProductPickerPropertyEditor.cs) for the Shopify property editor equivalent.

src/Umbraco.Cms.Integrations.Crm.Dynamics/Umbraco.Cms.Integrations.Crm.Dynamics.csproj
AndyButland, line 1: Looks like this is missing the `<SupportedPlatform Include="browser" />` setting.  Suggest compare very carefully to:
https://github.com/umbraco/Umbraco.Cms.Integrations/blob/main-v14/src/Umbraco.Cms.Integrations.Commerce.Shopify/Umbraco.Cms.Integrations.Commerce.Shopify.csproj

src/Umbraco.Cms.Integrations.Crm.Dynamics/wwwroot/umbraco-package.json
```diff
@@ -0,0 +1,14 @@
+{
+    "id": "Umbraco.Cms.Integrations.Crm.Dynamics",
+    "name": "Umbraco CMS Integrations: Crm - Dynamics",
+    "version": "1.0.0",
```
AndyButland, line 4: Should bump this version to whatever the next major is.

AndyButland (commented):
Looking good @dtng95.  I just cast an eye over progress so far and left you a few comments... hope they are useful.

dtng95:
Thanks @AndyButland for the above comments. I will check those out.

src/Umbraco.Cms.Integrations.Commerce.Shopify/Client/config.outputPath.js
acoumb: You can ignore this as I'll correct it for _Shopify_ through a different PR.

Issue #209, opened 2024-09-10
[Algolia integration] DateTime properties are not sortable due to the DateTime.ToString() call when building an index.


giuunit:
This issue is linked to the Umbraco.Cms.Integrations.Search.Algolia plugin. 

When calling the Search/BuildIndex endpoint to build an index based on doc types using the umbraco backoffice plugin, the generic date properties like createDate and updateDate are serialized into string using the DateTime.ToString() function. It is an issue as this function relies on the current culture of the server and doesn't allow for an easily sortable date to sort for algolia. 

Algolia suggests to convert the dates into Unix timestamps here: https://www.algolia.com/doc/guides/managing-results/refine-results/sorting/how-to/sort-an-index-by-date/#convert-dates-into-unix-timestamps-an-example to make them sortable. 

A pull request is on the way. Thank you.

---
_This item has been added to our backlog AB#43864_

acoumb:
Hi @giuunit ,

You PR #210 has been merged and released.

Thank you for your support.

PR #210, opened 2024-09-10
converted createdDate and updatedDate to unix timestamp

giuunit:
The issue was highlighted here: https://github.com/umbraco/Umbraco.Cms.Integrations/issues/209 

The recommendation from Algolia is here: https://www.algolia.com/doc/guides/managing-results/refine-results/sorting/how-to/sort-an-index-by-date/#convert-dates-into-unix-timestamps-an-example


AndyButland:
This looks like the right approach to me, but I'm just thinking about upgrades and breaking changes,  If someone has already indexed dates and relies on them as strings, this change could break their implementation.

So we should either consider this as a breaking change and document as such, or, perhaps better - we add these as additional fields rather than replacing the existing ones.  So we'd end up with an index with `CreateDate` and `CreateDateTimestamp` - where you could sort on one, and if you are already using the other, it won't be affected.

Thoughts @acoumb?

acoumb:
I agree with @AndyButland on this one, and I see Algolia uses this path, to keep the actual date as a string representation, and use an additional [timestamp](https://www.algolia.com/doc/guides/managing-results/refine-results/sorting/how-to/sort-an-index-by-date/#after).

Since this is the first time the situation presented, I would presume that introducing this breaking change could affect a lot of users working with Algolia.

giuunit:
Hi gents. You are absolutely correct. I will update my MR and add new timestamp properties to the class rather than updating the current properties. The update should arrive shortly.

AndyButland:
Nice, thanks very much @giuunit for the contribution.  We'll see about getting this out in a release shortly.

giuunit:
It's my pleasure, it's fixing an issue regarding a project for a client I am currently working on so it is benefiting me as well.

PR #212, opened 2024-09-16
MNTP index value converter + fixing issue in helper function

giuunit:
There was an issue with the TryGetPropertyIndexValue where we could still get a NullRef exception for some edge cases that is fixed now. 

I also added a Index value converter for MNTP based on the name that use cache and db call when needed. As it can impact the database I left it outside of the list of default converters so people can decide to pick it or not. Tested in v10.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Search.Algolia/Converters/SimpleMultinodeTreePickerIndexValueConverter.cs
+++ b/src/Umbraco.Cms.Integrations.Search.Algolia/Converters/SimpleMultinodeTreePickerIndexValueConverter.cs
@@ -1,0 +1,68 @@
+﻿using Umbraco.Cms.Core;
+using Umbraco.Cms.Core.Models;
+using Umbraco.Cms.Core.Services;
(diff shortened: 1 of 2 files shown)


giuunit:
Hi, is there an opportunity to review my PR ? thank you. It looks like some checks are not successful but I do not believe it is related to my changes. Can someone re run them ?

acoumb:
Hi @giuunit ,

The integration comes with a set of default converters but we try to limit that as much as possible, allowing instead for users to register their own as detailed [here](https://docs.umbraco.com/umbraco-dxp/integrations/algolia#version-2.1.5-and-up) (or replace one of the existing ones).

So I would recommend you register your `SimpleMultinodeTreePickerIndexValueConverter` on your side. Adding too many converters by default, for even more complex types in Umbraco, might not suit with everyone using the integration. So I think it's best to keep the scopes isolated.

I will release today patch versions with the fix for the `ConverterExtensions` helper.

Thank you,
Adrian

giuunit:
thats awesome thank you

PR #214, opened 2024-09-18
Fix NULL reference

acoumb:
Current PR adds a fix to the converter extension method following a potential NULL reference error reported in #212 .


src/Umbraco.Cms.Integrations.Search.Algolia/Extensions/ConverterExtensions.cs
```diff
@@ -13,8 +13,7 @@ public static bool TryGetPropertyIndexValue(this IProperty property, out string
             {
                 success = false;
             }
-
-            value = property.GetValue().ToString();
+            else value = property.GetValue().ToString();
```
AndyButland, line 16:
```suggestion
            else
            {
                value = property.GetValue().ToString();
            }
```

AndyButland (approved):
Just a code styling suggestion to use braces, otherwise looks like a good catch.

Issue #213, opened 2024-09-18
[Algolia integration] the current project doesn't allow geolocation indexing


giuunit:
Based on Algolia documentation, we should be able to add geolocation properties inside algolia using the format listed here: 

https://www.algolia.com/doc/guides/managing-results/refine-results/geolocation/#enabling-geo-search-by-adding-geolocation-data-to-records

We need a record with a "_geoloc" attribute that contains an object with lat/long properties. 

The issue we have is that all the "hardcoded" properties are locked inside the SearchController.

I think it would be good to have a "before indexing" event for the Record object where we would be able to edit any property like non data properties after the hydration from the converters based on the Data dictionary.

giuunit:
Here's the code I have so far: 

```
using Newtonsoft.Json;

namespace Umbraco.Cms.Integrations.Search.Algolia.Models
{
    public class GeolocProperties
    {
        [JsonProperty("lat")]
        public double Latitude { get; set; }

        [JsonProperty("lng")]
        public double Longitude { get; set; }
    }
}
```

```
using Newtonsoft.Json;

namespace Umbraco.Cms.Integrations.Search.Algolia.Models
{
    public class Record
    {
        public Record()
        {
            Data = new Dictionary<string, object>();
            Geoloc = new List<GeolocProperties>();
        }

        public Record(Record record)
        {
            ContentTypeAlias = record.ContentTypeAlias;
            ObjectID = record.ObjectID;
            Id = record.Id;
            Name = record.Name;
            CreateDate = record.CreateDate;
            CreatorName = record.CreatorName;
            UpdateDate = record.UpdateDate;
            WriterName = record.WriterName;
            TemplateId = record.TemplateId;
            Level = record.Level;
            Path = record.Path;
            Url = record.Url;
            Data = record.Data;
            Geoloc = record.Geoloc;
        }

        public string ObjectID { get; set; }

        public int Id { get; set; }

        public string Name { get; set; }

        public string CreateDate { get; set; }

        /// <summary>
        /// Unix timestamp
        /// </summary>
        public long CreateDateTimestamp { get; set; }

        public string CreatorName { get; set; }

        public string UpdateDate { get; set; }

        /// <summary>
        /// Unix timestamp
        /// </summary>
        public long UpdateDateTimestamp { get; set; }

        public string WriterName { get; set; }

        public int TemplateId { get; set; }

        public int Level { get; set; }

        public List<string> Path { get; set; }

[...]

PR #215, opened 2024-09-18
geo loc properties

giuunit:
Based on the issue here: https://github.com/umbraco/Umbraco.Cms.Integrations/issues/213

Based on the geo location feature in Algolia, we need a root level property _geoloc to contain geo location data to enable geo location search. It cannot be inside the Data property as it is a simple Dictionary<string, object> with no way to tell that the data needs to be at the root level (all the properties inside the Data dictionary are serialized as data.prop)

https://www.algolia.com/doc/guides/managing-results/refine-results/geolocation/#enabling-geo-search-by-adding-geolocation-data-to-records


giuunit:
Many thanks for approving it with such short notice. I understand it is not perfect - this approach would heavily relies on the IRecordBuilder and doesn't work "as is" but for my client it will have a huge impact. I would love to help more regarding a more robust integration, my email is: giuliano.dore@gmail.com. Thank you

PR #216, opened 2024-09-18
Add Algolia geolocation service

acoumb:
Current PR adds support for geolocation in Algolia following these discussions: #213 and #215.

This update adds the new property for managing multiple coordinates - I've renamed the changes from #215 so they would be more consistent:
* `GeolocProperties` = `GeolocationEntity`
* `Record.Geoloc` = `Record.Geolocation`

Responsibility for managing geolocation data will be reserved to users interested in this feature, by registering a custom provider that implements `IAlgoliaGeolocationProvider`. This uses only an `async` method as I considered that usually coordinates are retrieved from external APIs using HTTP request. The provider will need to be registered in the composer as Singleton or Transient  to avoid exceptions from services invoking it. 

The registered provider will be picked by `AlgoliaGeolocationService` and the values used on indexing.

After wrapping this up, I will make updates to docs to provide details on the matter, and then to `v14` version of the package.

Current NuGet package version is `2.2.1` and this one I bumped to `2.3.0`.


AndyButland (commented):
Wondering if you have a bit more abstraction than is necessary here.  Could you consider:

- Register a singleton implementation if `IAlgoliaGeoLocationProvider` called `NullAlgoliaGeoLocationProvider` and have that implementation just return a null result.
- Inject that into `ContentRecordBuilder and use it.
- An implementor would register and replace the implementation with their own in order to use it to get anything other than null results.

To support that the property on `Record` should be nullable, such that if there's no geolocation nothing is indexed in Algolia (not even an empty array.  But if for some reason you can't make it nullable, your default implementation can return an empty array instead of null, and be renamed to `EmptyAlgoliaGeoLocationProvider`

acoumb:
Updated

AndyButland (commented):
On reflection, maybe the nullability is a bit weird now.  It's generally good practice for methods returning collections to return empty collections rather than null.  So I'd suggest updating `IAlgoliaGeoLocationProvider` to return not null and the default implementation to return an empty collection.

Then for `Record` - maybe test with sending an empty array.  If that seems OK in Algolia you can remove the nullability of that too.  If it presents issues, then keep this as null and set to null if you get an empty collection back from the geo-location provider, otherwise set it to the collection.

How does that sound?

acoumb:
On the default `AlgoliaNullGeolocationProvider` that returns `NULL`, the `_geoloc` property is skipped.

AndyButland (approved):

PR #218, opened 2024-09-19
Pass content item to geolocation method

acoumb:
This PR is a follow up to [this](https://github.com/umbraco/Umbraco.Cms.Integrations/issues/213#issuecomment-2360663317) comment, realizing that another use case I skipped was the one where the localization information might come from content.

To address this, I am passing an `IContent`  parameter to the provider method.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaIndexService.cs
+++ b/src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaIndexService.cs
@@ -32,7 +32,7 @@ namespace Umbraco.Cms.Integrations.Search.Algolia.Services
                     : new List<Record> {
                         new Record {
                             ObjectID = Guid.NewGuid().ToString(),
-                            GeolocationData = await _algoliaGeolocationProvider.GetGeolocationAsync(),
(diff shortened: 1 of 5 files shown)


src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaIndexService.cs
```diff
@@ -32,7 +32,7 @@ await index.SaveObjectsAsync(payload != null
                     : new List<Record> {
                         new Record {
                             ObjectID = Guid.NewGuid().ToString(),
-                            GeolocationData = await _algoliaGeolocationProvider.GetGeolocationAsync(),
+                            GeolocationData = new List<GeolocationEntity>(),
```
AndyButland, line 35: null rather than empty collection?

AndyButland (approved):

src/Umbraco.Cms.Integrations.Search.Algolia/Services/AlgoliaIndexService.cs
acoumb: Updated

PR #221, opened 2024-10-10
V14/feature/43247 v14 integrations (semrush)

dtng95:
- This is the implementation of Semrush Integration. Below is the image of the Semrush workspace. 
image

- Check these links in order to test with Semrush: 
https://docs.umbraco.com/umbraco-dxp/integrations/semrush
https://umbraco.com/blog/integrating-umbraco-cms-with-semrush/

- There are still some namespaces that need to be cleaned up, will be updated in the next PR update.


src/Umbraco.Cms.Integrations.Testsite.V14/Umbraco.Cms.Integrations.Testsite.V14.csproj
```diff
@@ -12,11 +20,7 @@
   <ItemGroup>
     <!-- Opt-in to app-local ICU to ensure consistent globalization APIs across different platforms -->
     <PackageReference Include="Microsoft.ICU.ICU4C.Runtime" Version="72.1.0.3" />
-    <ProjectReference Include="..\Umbraco.Cms.Integrations.Crm.ActiveCampaign\Umbraco.Cms.Integrations.Crm.ActiveCampaign.csproj" />
-    <ProjectReference Include="..\Umbraco.Cms.Integrations.Crm.Dynamics\Umbraco.Cms.Integrations.Crm.Dynamics.csproj" />
-    <ProjectReference Include="..\Umbraco.Cms.Integrations.Automation.Zapier\Umbraco.Cms.Integrations.Automation.Zapier.csproj" />
-    <ProjectReference Include="..\Umbraco.Cms.Integrations.Crm.Hubspot\Umbraco.Cms.Integrations.Crm.Hubspot.csproj" />
-    <ProjectReference Include="..\Umbraco.Cms.Integrations.Search.Algolia\Umbraco.Cms.Integrations.Search.Algolia.csproj" />
+    <ProjectReference Include="..\Umbraco.Cms.Integrations.SEO.Semrush\Umbraco.Cms.Integrations.SEO.Semrush.csproj" />
```
dtng95, line 23: - I was wondering if we could remove these references in the test site, and later if someone want to test a package or anything in this repo, they can add the reference by themselves.

src/Umbraco.Cms.Integrations.Testsite.V14/Umbraco.Cms.Integrations.Testsite.V14.csproj
```diff
@@ -4,6 +4,14 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
   </PropertyGroup>
+  <ItemGroup>
```
dtng95, line 7: I include these 2 files `semrushColumns.json` and `semrushDataSources.json` in the test site, because I'm not sure where to put them properly. wwwroot or in `Umbraco.Cms.Integrations.SEO.Semrush` project

src/Umbraco.Cms.Integrations.SEO.Semrush/SemrushContentApp.cs
```diff
@@ -1,37 +1,30 @@
-﻿using System.Collections.Generic;
+﻿//using System.Collections.Generic;
```
dtng95, line 1: Will remove in the next update.

src/Umbraco.Cms.Integrations.SEO.Semrush/SemrushComposer.cs
```diff
@@ -1,32 +1,25 @@
...
-#if NETCOREAPP
         public void Compose(IUmbracoBuilder builder)
         {
             var options = builder.Services.AddOptions<SemrushSettings>()
                 .Bind(builder.Config.GetSection(Constants.Configuration.Settings));
             var oauthOptions = builder.Services.AddOptions<SemrushOAuthSettings>()
                 .Bind(builder.Config.GetSection(Constants.Configuration.OAuthSettings));
 
-            builder.ContentApps().Append<SemrushContentApp>();
+            //builder.ContentApps().Append<SemrushContentApp>();
```
dtng95, line 22: Will remove in the next update

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/GetCurrentContentPropertiesController.cs
```diff
@@ -0,0 +1,44 @@
...
+using Umbraco.Cms.Core.Services;
+using Umbraco.Cms.Integrations.SEO.Semrush.Configuration;
+using Umbraco.Cms.Integrations.SEO.Semrush.Models.Dtos;
+using Umbraco.Cms.Integrations.SEO.Semrush.Services;
+
+namespace Umbraco.Cms.Integrations.SEO.Semrush.Api.Management.Controllers
+{
+    [ApiVersion("1.0")]
+    [ApiExplorerSettings(GroupName = Constants.ManagementApi.SemrushGroupName)]
+    public class GetCurrentContentPropertiesController : SemrushControllerBase
```
dtng95, line 15: This is the alternate approach that I use in order to get properties and their values of current content and display it on the UI. Since CMS team have proposed another approach to get the properties in client, I will remove this in the next PR update.

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/PingController.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Umbraco.Cms.Integrations.SEO.Semrush.Configuration;
+using Umbraco.Cms.Integrations.SEO.Semrush.Services;
+
+namespace Umbraco.Cms.Integrations.SEO.Semrush.Api.Management.Controllers
+{
+    [ApiVersion("1.0")]
```
dtng95, line 16: Do we need this endpoint?

src/Umbraco.Cms.Integrations.Testsite.V14/Umbraco.Cms.Integrations.Testsite.V14.csproj
AndyButland: @acoumb I don't remember what the reason for these files were... but if it was to allow users to update them, that won't be possible now they are coming from a Razor Class Library.  So maybe their contents should be removed and added as constants in the code (if there's no need for customers to update), or as configuration (if there is).

src/Umbraco.Cms.Integrations.Testsite.V14/Umbraco.Cms.Integrations.Testsite.V14.csproj
AndyButland: Maybe leave them in but commented out?  That way it's easy to add the one you need for testing.

src/Umbraco.Cms.Integrations.Testsite.V14/Umbraco.Cms.Integrations.Testsite.V14.csproj
AndyButland: I see they are quite big though... so not ideal in configuration or constants.  Perhaps embedded resources is the right way to go, used as a fallback having first checked for a file on disk if the customer has provided it?

AndyButland (commented):
Generally looks good to me @dtng95.  Have left a few comments but will leave now for @acoumb to take it for a spin when he has time.  Thanks.

src/Umbraco.Cms.Integrations.SEO.Semrush/Client/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "lit": "^3.2.0"
+  },
+  "devDependencies": {
+    "typescript": "^5.5.3",
+    "vite": "^5.4.1",
+    "vite-tsconfig-paths": "^5.0.1",
+    "@hey-api/openapi-ts": "^0.46.3",
+    "@typescript-eslint/eslint-plugin": "^7.13.0",
+    "@typescript-eslint/parser": "^7.13.0",
+    "@umbraco-cms/backoffice": "^14.3.0",
```
AndyButland, line 26: If you use 14,3 here, we should also make sure we add that as the minimum version for the NuGet dependency too (i.e. 14.3.0 rather than 14.0.0).  That's fine to do, but just want to make sure we have these aligned.

dtng95: I'm having issue while getting content properties using UMB_CONTENT_WORKSPACE_CONTEXT, it returns undefined. Niels suggest that I should use cms version 15 at least. Should I use version 15 in this integration package?

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/GetCurrentContentPropertiesController.cs
```diff
@@ -0,0 +1,44 @@
...
+        [HttpGet("content-properties")]
+        [ProducesResponseType(typeof(List<ContentPropertyDto>), StatusCodes.Status200OK)]
+        public async Task<IActionResult> GetCurrentContentProperties(string contentId)
+        {
+            var propertyList = new List<ContentPropertyDto>();
+
+            var content = _contentService.GetById(Guid.Parse(contentId));
+            if (content != null)
+            {
+                propertyList = content.Properties.Select(s => new ContentPropertyDto { PropertyName = s.PropertyType.Name, PropertyValue = s.Values.ToList()[0].PublishedValue.ToString() }).ToList();
```
AndyButland, line 38: Probably this should be a bit more "defensive".  E.g. are we sure `s.Values` couldn't be null?  Or could be a collection with zero entries?  In those cases we'd get an exception here.

src/Umbraco.Cms.Integrations.SEO.Semrush/Client/package.json
AndyButland: I just responded in Slack... but to repeat so we have it here, let's stick with what you have 
 - it's an extra server-trip but unlikely to be a noticeable delay since we are calling Semrush's API in this integration too.  And means we can align with getting all the integrations we plan to release out on 14.
 
Maybe when 15 is out we can look again and release an update that has a minimum dependency on that, and includes this optimization of avoiding a server round-trip.

src/Umbraco.Cms.Integrations.Testsite.V14/Umbraco.Cms.Integrations.Testsite.V14.csproj
acoumb: My thought is that you can leave them all referenced, as with the completion of Semrush and merge into main, when someone would want to play with the integrations they would have the web environment ready to go.

src/Umbraco.Cms.Integrations.Testsite.V14/Umbraco.Cms.Integrations.Testsite.V14.csproj
acoumb: The files are used to display the tooltips in the results table. So you can either use them from the JSON files, load them server-side or cache them.

acoumb: I picked the data from their API [docs](https://developer.semrush.com/api/v3/analytics/basic-docs/).

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/Token/ValidateTokenController.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+        }
+
+        [HttpGet("validate")]
+        [ProducesResponseType(typeof(AuthorizationResponseDto), StatusCodes.Status200OK)]
+        public async Task<IActionResult> ValidateToken()
+        {
+            _semrushTokenService.TryGetParameters(Constants.TokenDbKey, out TokenDto token);
+
+            if (!token.IsAccessTokenAvailable) return Ok(new AuthorizationResponseDto { IsAuthorized = false });
```
acoumb, line 25: With the new API endpoints pattern, I think it would make sense and be logical to return an `Unauthorized` response.

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/Token/ValidateTokenController.cs
```diff
@@ -0,0 +1,46 @@
...
+
+            var response = await ClientFactory()
```
acoumb, line 27: We should probably remove this and use the `IHttpClientFactory` approach. And anywhere else in the Semrush project.

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/GetRelatedPhrasesController.cs
```diff
@@ -0,0 +1,72 @@
...
+            var response = await ClientFactory()
+                .GetAsync(string.Format(Constants.SemrushKeywordsEndpoint, _settings.BaseUrl, method, token.AccessToken, phrase, dataSource));
+
+            if (response.IsSuccessStatusCode)
+            {
+                var responseContent = await response.Content.ReadAsStringAsync();
+
+                var relatedPhrasesDeserialized = JsonSerializer.Deserialize<RelatedPhrasesDto>(responseContent);
+
+                if (!relatedPhrasesDeserialized.IsSuccessful) return Ok(relatedPhrasesDeserialized);
```
acoumb, line 56: On my SEMrush account I got this error: `ERROR 134 :: TOTAL LIMIT EXCEEDED` on search, but the UI did not reflect this. Could you add some error handling for this use case?

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/Token/GetTokenDetailsController.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Umbraco.Cms.Integrations.SEO.Semrush.Api.Management.Controllers.Token
+{
+    public class GetTokenDetailsController : TokenControllerBase
+    {
+        public GetTokenDetailsController(IOptions<SemrushSettings> options, IWebHostEnvironment webHostEnvironment, ISemrushTokenService semrushTokenService, ICacheHelper cacheHelper, TokenBuilder tokenBuilder, SemrushComposer.AuthorizationImplementationFactory authorizationImplementationFactory) : base(options, webHostEnvironment, semrushTokenService, cacheHelper, tokenBuilder, authorizationImplementationFactory)
+        {
+        }
+
+        [HttpGet("detail")]
```
acoumb, line 23: Could you rename this to `details`?

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/Token/GetAccessTokenController.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace Umbraco.Cms.Integrations.SEO.Semrush.Api.Management.Controllers.Token
+{
+    public class GetAccessTokenController : TokenControllerBase
+    {
+        public GetAccessTokenController(IOptions<SemrushSettings> options, IWebHostEnvironment webHostEnvironment, ISemrushTokenService semrushTokenService, ICacheHelper cacheHelper, TokenBuilder tokenBuilder, SemrushComposer.AuthorizationImplementationFactory authorizationImplementationFactory) : base(options, webHostEnvironment, semrushTokenService, cacheHelper, tokenBuilder, authorizationImplementationFactory)
+        {
+        }
+
+        [HttpPost("get")]
+        [ProducesResponseType(typeof(string), StatusCodes.Status200OK)]
```
acoumb, line 24: Could consider also a `500-Internal Server Error` as the service method could encounter an error. So you could validate the returned string if it contains "Error" and handle accordingly.

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/Token/RefreshAccessTokenController.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace Umbraco.Cms.Integrations.SEO.Semrush.Api.Management.Controllers.Token
+{
+    public class RefreshAccessTokenController : TokenControllerBase
+    {
+        public RefreshAccessTokenController(IOptions<SemrushSettings> options, IWebHostEnvironment webHostEnvironment, ISemrushTokenService semrushTokenService, ICacheHelper cacheHelper, TokenBuilder tokenBuilder, SemrushComposer.AuthorizationImplementationFactory authorizationImplementationFactory) : base(options, webHostEnvironment, semrushTokenService, cacheHelper, tokenBuilder, authorizationImplementationFactory)
+        {
+        }
+
+        [HttpPost("refresh")]
+        [ProducesResponseType(typeof(string), StatusCodes.Status200OK)]
```
acoumb, line 23: Similar to a comment above, might consider handling `500` as the service could return a string with `error` message.

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/GetAuthorizationUrlController.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    [ApiVersion("1.0")]
+    [ApiExplorerSettings(GroupName = Constants.ManagementApi.SemrushGroupName)]
+    public class GetAuthorizationUrlController : SemrushControllerBase
+    {
+        public GetAuthorizationUrlController(IOptions<SemrushSettings> options, IWebHostEnvironment webHostEnvironment, ISemrushTokenService semrushTokenService, ICacheHelper cacheHelper, TokenBuilder tokenBuilder, SemrushComposer.AuthorizationImplementationFactory authorizationImplementationFactory) : base(options, webHostEnvironment, semrushTokenService, cacheHelper, tokenBuilder, authorizationImplementationFactory)
+        {
+        }
+
+        [HttpGet("auth-url")]
```
acoumb, line 24: Could we use the path `auth/url`?

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/GetColumnsController.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        public GetColumnsController(IOptions<SemrushSettings> options, IWebHostEnvironment webHostEnvironment, ISemrushTokenService semrushTokenService, ICacheHelper cacheHelper, TokenBuilder tokenBuilder, SemrushComposer.AuthorizationImplementationFactory authorizationImplementationFactory) : base(options, webHostEnvironment, semrushTokenService, cacheHelper, tokenBuilder, authorizationImplementationFactory)
+        {
+        }
+
+        [HttpGet("columns")]
+        [ProducesResponseType(typeof(IEnumerable<ColumnDto>), StatusCodes.Status200OK)]
+        public IActionResult GetColumns()
+        {
+            string semrushColumnsPath = $"{_webHostEnvironment.ContentRootPath}/App_Plugins/UmbracoCms.Integrations/SEO/Semrush/semrushColumns.json";
```
acoumb, line 30: This would need to be updated depending on the new path.

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/GetDataSourcesController.cs
```diff
@@ -0,0 +1,69 @@
...
+{
+    [ApiVersion("1.0")]
+    [ApiExplorerSettings(GroupName = Constants.ManagementApi.SemrushGroupName)]
+    public class GetDataSourcesController : SemrushControllerBase
+    {
+        public GetDataSourcesController(IOptions<SemrushSettings> options, IWebHostEnvironment webHostEnvironment, ISemrushTokenService semrushTokenService, ICacheHelper cacheHelper, TokenBuilder tokenBuilder, SemrushComposer.AuthorizationImplementationFactory authorizationImplementationFactory) : base(options, webHostEnvironment, semrushTokenService, cacheHelper, tokenBuilder, authorizationImplementationFactory)
+        {
+        }
+
+        [HttpGet("datasources")]
```
acoumb, line 26: Rename to `data-sources`

acoumb:
A couple of items I've noticed during testing, grouped by functional tests and implementation.
### Functional
* If my content node has multiple tabs with properties, the API throws an exception and the properties are not added to the select control.
image
This is my content structure:
image
image
The error occurs only with tabbed content types.
image

* Any possibility to group the properties by their group title? For example:
image

I see that the [`uui-select`](https://uui.umbraco.com/?path=/docs/uui-select--docs) component allows it.
image

* Data sources and methods controls tooltips are not being displayed
image

* This section should be hidden on first load and display when a property is selected or on _Search new_ action
image

* Could we keep the _Revoke_ button red, for consistency with it's purpose?
image

* Tooltips for results columns do not show up
image

* When the _Search keywords_ action is triggered, no loader is displayed. If I perform multiple consecutive searches, until the results table is updated I do not know what is happening.

* The path has an extra `/` before semrush
image

### Implementation
* Namespaces cleanup
* Since there are a lot of endpoints and one might be confused regarding their purpose, could you restructure the `Access Token` ones to have the following pattern: `/umbraco/semrush/management/api/v1/{operation}`. I think this way they would provide more context towards their use case.
image
* `Configuration | SemrushOAuthSettings`: you can remove the constructor that takes `NameValueCollection` as parameter, it was used for v8 version of the integration.
*  `Configuration | SemrushSettings`: you can remove the constructor that takes `NameValueCollection` as parameter, it was used for v8 version of the integration.
* `Constants | Configuration`: you can remove the properties with value _Umbraco.Cms.Integrations.SEO.Semrush.***_, they were used for v8 version of the integration.

acoumb (changes requested):
Dropped some feedback in a comment, @AndyButland if you care to look them up, maybe I got something wrong.
Thanks

AndyButland:
All looks sensible to me @acoumb, thanks.

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/GetRelatedPhrasesController.cs
dtng95: Already added to display the message, will be included in the next PR.

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/Token/ValidateTokenController.cs
dtng95: I decided to not return `Unauthorized`, because it will force user to login again, and again because it keeps returning `Unauthorized` if `token.IsAccessTokenAvailable` is `false`

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/GetCurrentContentPropertiesController.cs
```diff
@@ -0,0 +1,66 @@
...
+            _contentService = contentService;
+            _contentTypeService = contentTypeService;
+        }
+
+        [HttpGet("content-properties")]
+        [ProducesResponseType(typeof(List<ContentPropertyDto>), StatusCodes.Status200OK)]
+        public async Task<IActionResult> GetCurrentContentProperties(string contentId)
+        {
+            var propertyList = new List<ContentPropertyDto>();
+            var currentContent = _contentService.GetById(Guid.Parse(contentId));
```
dtng95, line 38: I had to use both `IContentService` and `IContentTypeService` here because `IContentTypeService` can get properties of a content with its group, but cannot get their values. That's why I also used `IContentService` to get their corresponding values.

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/PingController.cs
acoumb: No, this can be removed.

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/GetCurrentContentPropertiesController.cs
AndyButland: Would suggest `contentId` can be passed as a GUID, which will save parsing it for the call to get it by Id.

Need to check here too if this logic works with compositions - e.g. content type A as a property, and composition of content type B that also has a property.  I think you'll only get the property from content type A here.

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/SemrushControllerBase.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace Umbraco.Cms.Integrations.SEO.Semrush.Api.Management.Controllers
+{
+    [ApiController]
+    [BackOfficeRoute($"{Constants.ManagementApi.RootPath}/v{{version:apiVersion}}")]
+    [Authorize(Policy = AuthorizationPolicies.BackOfficeAccess)]
+    [MapToApi(Constants.ManagementApi.ApiName)]
+    public class SemrushControllerBase : Controller
+    {
+        protected static IHttpClientFactory _clientFactory;
+        private IOptions<SemrushSettings> options;
```
acoumb, line 21: Some of these members are not used anymore, you can remove them.

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/Token/GetAccessTokenController.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+        }
+
+        [HttpPost("token/get")]
+        [ProducesResponseType(typeof(string), StatusCodes.Status200OK)]
+        [ProducesResponseType(typeof(string), StatusCodes.Status500InternalServerError)]
+        public async Task<IActionResult> GetAccessToken([FromBody] AuthorizationRequestDto request)
+        {
+            var result = await _authorizationService.GetAccessTokenAsync(request.Code);
+            return result.StartsWith("Error") ? BadRequest(result) : Ok(result);
```
acoumb, line 30: Could you update to `InternalServerError` to align with the `ProductResponseType` attribute status code?

src/Umbraco.Cms.Integrations.SEO.Semrush/Api/Management/Controllers/Token/RevokeTokenController.cs
```diff
@@ -0,0 +1,32 @@
...
+            IWebHostEnvironment webHostEnvironment,
+            ISemrushTokenService semrushTokenService,
+            ICacheHelper cacheHelper,
+            TokenBuilder tokenBuilder,
+            SemrushComposer.AuthorizationImplementationFactory authorizationImplementationFactory,
+            IHttpClientFactory httpClientFactory) : base(options, webHostEnvironment, semrushTokenService, cacheHelper, tokenBuilder, authorizationImplementationFactory, httpClientFactory)
+        {
+        }
+
+        [HttpPost("token/revoke")]
```
acoumb, line 22: Could you add `ProducesResponseType` for `200` ?

acoumb (changes requested):
Looks good to me, left a few comments inline for controllers, and the two below.

* Could you remove the `ClientFactory` pattern in services and replace them with `IHttpClientFactory`?

* And if you could central align the loader in these views:
image
image

Following these I'm all good with merging them in. Thanks.

After

src/Umbraco.Cms.Integrations.SEO.Semrush/Services/BaseAuthorizationService.cs
```diff
@@ -10,20 +10,17 @@ public class BaseAuthorizationService
             "&scope={3}";
 
         // Using a static HttpClient (see: https://www.aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/).
```
acoumb, line 12: You can remove this comment, it is not relevant anymore.

acoumb (approved):
Looks good. Good to go with merging it in and tests with the `*.nupkg` version.

Issue #222, opened 2024-10-21
Migration of old indicies to Umbraco 14


evaHopNltg:
When we upgrade AlgoliaSearch, from 13 to 14, our old indicies doesn't work in the dashboard beacause the datamodel have changed. 

When I look in the migration script, the only concern are if the table exists or not. 

---
_This item has been added to our backlog AB#45268_

acoumb:
Hi @evaHopNltg ,
I've released version [`3.2.0`](https://www.nuget.org/packages/Umbraco.Cms.Integrations.Search.Algolia/3.2.0) of Algolia, which addresses this issue.

Regards,
Adrian

PR #224, opened 2024-11-04
Add migration for updating an index serialized data following Umbraco…

acoumb:
Current PR fixes the issue raised with #222, where after upgrading from version 13 to 14, the existing indices are not working.
The reason for this is the difference between the structure of the indices stored in the `algoliaIndices` table, that has changed between the two versions.

While the version for 13 had this structure serialized into the database,
```
[{
	"contentType": {
		"alias": "home",
		"name": "Home",
		"icon": "icon-home"
	},
	"properties": [{
			"alias": "title",
			"name": "Title",
			"icon": null
		}
	],
	"propertiesDescription": ["Title"]
}]
```
with version 14, the new structure was this one:
```
[{
	"Id": 1056,
	"Icon": "icon-home",
	"Alias": "home",
	"Name": "Home",
	"Selected": true,
	"AllowRemove": false,
	"Properties": [{
	        "Id": 52,
	        "Icon": "icon-document",
	        "Alias": "title",
	        "Name": "Title",
	        "Group": "Properties",
	        "Selected": true
        }]
}]
```

This resulted in inconsistencies of the indices following an upgrade.

To address this I've made the following changes:
* Added a property called `ContentType` to match the property from old indices.
* Added the `UpdateAlgoliaIndicesPostUpgrade` migration that will look up the database indices, and in case it finds one with the `ContentType` property - meaning it is an index from v13 - it will update the serialized data to match the current index structure.
* Additional `JsonPropertyName` attributes to map the properties of the content type dto.

Update will go under a new minor version - `3.2.0`.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Search.Algolia/Migrations/UpdateAlgoliaIndicesPostUpgrade.cs
+++ b/src/Umbraco.Cms.Integrations.Search.Algolia/Migrations/UpdateAlgoliaIndicesPostUpgrade.cs
@@ -1,4 +1,5 @@
 ﻿using System.Text.Json;
+using System.Text.Json.Serialization;
 using Umbraco.Cms.Core.Services;
 using Umbraco.Cms.Infrastructure.Migrations;
 using Umbraco.Cms.Infrastructure.Scoping;
@@ -11,7 +12,7 @@ namespace Umbraco.Cms.Integrations.Search.Algolia.Migrations
         private readonly IScopeProvider _scopeProvider;
         private readonly IContentTypeService _contentTypeService;
 
-        public UpdateAlgoliaIndicesPostUpgrade(IMigrationContext context, IScopeProvider scopeProvider, IContentTypeService contentTypeService) 
+        public UpdateAlgoliaIndicesPostUpgrade(IMigrationContext context, IScopeProvider scopeProvider, IContentTypeService contentTypeService)
             : base(context)
         {
             _scopeProvider = scopeProvider;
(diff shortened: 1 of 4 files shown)


AndyButland (commented):
It looks like you've had to hack this a little bit by adding properties to `ContentTypeDto` you don't really need moving forward.  I think the migration is the right way to go, but it might be cleaner to:

- Create a private class in the migration class, that represents the serialized data from the previous version.
- When looping through the indices, try deserializing to that.
- If you can - and hence you know you have one in the old format - create a new instance of the `ContentTypeDto` that has just the properties you need now.
- Save that back to the database.

acoumb:
Thanks @AndyButland , it makes sense to have the upgrade logic in a single place. Moved the upgrade feature into the migration plan, and left `ContentTypeDto` and `ContentTypePropertyDto` clean.

src/Umbraco.Cms.Integrations.Search.Algolia/Models/ContentTypeDtos/ContentTypeDto.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace Umbraco.Cms.Integrations.Search.Algolia.Models.ContentTypeDtos
+﻿using System.Text.Json.Serialization;
```
AndyButland, line 1: Probably unnecessary now.

src/Umbraco.Cms.Integrations.Search.Algolia/Migrations/UpdateAlgoliaIndicesPostUpgrade.cs
```diff
@@ -0,0 +1,95 @@
...
+        private class AlgoliaMigrationIndexData
+        {
+            [JsonPropertyName("contentType")]
+            public AlgoliaMigrationIndexContentType ContentType { get; set; }
+
+            [JsonPropertyName("properties")]
+            public IEnumerable<AlgoliaMigrationIndexProperty> Properties { get; set; }
+        }
+
+        public class AlgoliaMigrationIndexContentType
```
AndyButland, line 71: Should be `private` I think (and the class that follows).

AndyButland (approved):
Looks good to me now, thanks.

PR #227, opened 2024-11-13
Fix Shopify pagination selection persistence

acoumb:
Current PR addresses a couple of issues raised by the training team working with the Shopify integration, and a few minor fixes I've found during testing.

1. NULL reference exception when getting the page info details from the Shopify response
When there are few products, Shopify does not send the page info details with the response, therefor an exception was thrown at [this](https://github.com/umbraco/Umbraco.Cms.Integrations/blob/a40d8ca1ee088c168c6aed806d523b573d7b5e92/src/Umbraco.Cms.Integrations.Commerce.Shopify/Services/ShopifyService.cs#L149) line. On ready for release, I will also port this to versions of Shopify up to v13.
The fix meant adding a NULL check for the `pageInfoDetails` object.
2. Removing products from property editor throws a non-configurable error
shopify-delete
The `splice` method throws the above error, so as a workaround I thought of creating a copy of the products without the deleted product instead: https://github.com/umbraco/Umbraco.Cms.Integrations/blob/a40d8ca1ee088c168c6aed806d523b573d7b5e92/src/Umbraco.Cms.Integrations.Commerce.Shopify/Client/src/property-editor/shopify-product-picker-property-editor.element.ts#L126
3. Persisting selection across table pages
Products selected on one page would not be memorized when navigating other pages of the products list. The two objects that store the selections, 
    * `_selectionIdList` that would store an array of product IDs and
    * `_modalSelectedProducts` that store the products DTOs and send them back to the property editor on submit,
would keep the selections of the current table view. After going to a new page, the selections would be cleared.

    The workaround for this that I've put in place involves using two objects of the same type as the ones above (`_selectedItems` and `_selectedProducts`) which will hold selections across pages. To handle products being deselected, before pushing the selections into the array, I am removing from it any items of the current table view that might exist.

### DIFF
--- a/src/Umbraco.Cms.Integrations.Commerce.Shopify/Client/src/property-editor/shopify-product-picker-property-editor.element.ts
+++ b/src/Umbraco.Cms.Integrations.Commerce.Shopify/Client/src/property-editor/shopify-product-picker-property-editor.element.ts
@@ -122,7 +122,6 @@ export class ShopifyProductPickerPropertyEditor extends UmbLitElement implements
     }
 
     deleteProduct(id: number) {
-        var index = this.products.map(p => p.id).indexOf(id);
         this.products = this.products.filter(product => product.id != id);
         this.value = JSON.stringify(this.products.map(product => product.id));
         this.dispatchEvent(new CustomEvent('property-value-change'));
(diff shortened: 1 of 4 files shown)


src/Umbraco.Cms.Integrations.Commerce.Shopify/Client/src/property-editor/shopify-product-picker-property-editor.element.ts
```diff
@@ -123,7 +123,7 @@ export class ShopifyProductPickerPropertyEditor extends UmbLitElement implements
 
     deleteProduct(id: number) {
         var index = this.products.map(p => p.id).indexOf(id);
-        this.products.splice(index);
+        this.products = this.products.filter(product => product.id != id);
```
AndyButland, line 126: This looks right to me.  If the collection is immutable you'll have to replace it with a new collection without the item you want to remove.

AndyButland (approved):
All looks good to me.

PR #232, opened 2024-11-22
Update Algolia for v15

acoumb:
Current PR adds the updates for Umbraco 15 for Algolia.
The main changes included:
* Dashboard registration using `UmbExtensionManifest`
* Updated converters as a result of changes made to the CMS property index value factory `IPropertyIndexValueFactory` that returns an `IEnumerable<IndexValue>`.

Currently the indexing of `Umbraco.MediaPicker3` property editor does not work on my side in Examine, therefor I was not able to test the converter data. I will continue checking for updates with the CMS team.

In the meantime, I'm leaving this as an inspiration for the next integrations scheduled for upgrade.


AndyButland (approved):
All looks good to me.

PR #235, opened 2024-11-27
46038 Update HubSpot (CMS) integration package to 15

dtng95:
- Update CMS refenrece to 15
- Update client package to use Backoffice 15


AndyButland (approved):
You're flying!  Looks good to me.

acoumb (approved):
Looks good to me, just two small observations to consider before merge.
* Update package version to `6.0.0` in `Client > public > umbraco-package.json`
* With version 15, there is a console warning, so could you update the extension type in `umbraco-package.json` to use `backofficeEntryPoint`?
```
The `entryPoint` extension-type is deprecated, please use the `backofficeEntryPoint`.`
```

Good work!

Issue #239, opened 2024-12-19
Issues with dynamics views and Umbraco RuntimeMode Production


EelmanE:
## Issue
For production website we use[ Runtimemode Production in Umbraco](https://docs.umbraco.com/umbraco-cms/fundamentals/setup/server-setup/runtime-modes). But if we use the Dynamics package we cannot, because the 2 views in [the App_Plugins folder](https://github.com/umbraco/Umbraco.Cms.Integrations/tree/main/src/Umbraco.Cms.Integrations.Crm.Dynamics/App_Plugins/UmbracoCms.Integrations/Crm/Dynamics/Render) are copied to the output directory and cannot be precompiled. The DynamicsFormV8 has namespaces in it that Umbraco v13 doesn't have. 

I believe [this files](https://github.com/umbraco/Umbraco.Cms.Integrations/blob/main/src/Umbraco.Cms.Integrations.Crm.Dynamics/build/Umbraco.Cms.Integrations.Crm.Dynamics.targets) triggers the copy action of both view to our project.

I think the possible solutions are either only copy the relevant view based on the Umbraco version or use Razor class libaries to embed the view. 

## Details
Umbraco version: v13.5.2
Package: Umbraco.Cms.Integrations.Crm.Dynamics v1.3.6

acoumb:
Hi @EelmanE ,

Issue has been addressed in version `1.3.7` of the integrations.

Regards,
Adrian

EelmanE:
Hi Adrian, 

Great to hear! Thanks for the update.

Regards Erik

PR #243, opened 2025-01-29
Update Shopify and Dynamics to exclude V8 rendering view.

acoumb:
Current PR addresses the issue raised with item #239 , for integrations that have a rendering view for Umbraco 8. 

While running in production runtime mode, due to the V8 view, a compilation error occurs.

To address this, I've moved the rendering view to a different folder and excluded it during copy if the targeting framework is not .NET 4.7.2.


src/Umbraco.Cms.Integrations.Commerce.Shopify/Constants.cs
```diff
@@ -15,7 +15,7 @@ public static class Constants
 
         public static class RenderingComponent
         {
-            public const string DefaultV8ViewPath = AppPluginFolderPath + "/Render/Products.cshtml";
+            public const string DefaultV8ViewPath = AppPluginFolderPath + "/RenderV8/Products.cshtml";
 
             public const string DefaultV9ViewPath = AppPluginFolderPath + "/Render/ProductsV9.cshtml";
```
ronaldbarendse, line 20: Would it make sense to change this path to also have the version in the folder name?
```suggestion
            public const string DefaultV9ViewPath = AppPluginFolderPath + "/RenderV9/Products.cshtml";
```

src/Umbraco.Cms.Integrations.Commerce.Shopify/build/Umbraco.Cms.Integrations.Commerce.Shopify.targets
```diff
@@ -4,9 +4,10 @@
 		<ShopifyPropertyEditorFiles>$(MSBuildThisFileDirectory)..\App_Plugins\UmbracoCms.Integrations\Commerce\Shopify\**\*.*</ShopifyPropertyEditorFiles>
 	</PropertyGroup>
 
-	<Target Name="CopyShopifyPropertyEditorAssets" BeforeTargets="Build">
+	<Target Condition=" '$(TargetFramework)' != 'net472' " Name="CopyShopifyPropertyEditorAssets" BeforeTargets="Build">
```
ronaldbarendse, line 7: Shouldn't this condition be on the `ShopifyPropertyEditorFiles` item group instead, since this target needs to run on both .NET Framework and modern .NET?

And instead of adding multiple includes/excludes, use this instead:
```
<ShopifyPropertyEditorFiles Include="$(ShopifyPropertyEditorFiles)" />
<ShopifyPropertyEditorFiles Condition="'$(TargetFramework)' != 'net472'" Remove="$(MSBuildThisFileDirectory)..\App_Plugins\UmbracoCms.Integrations\Commerce\Shopify\RenderV8\*.*" />
<ShopifyPropertyEditorFiles Condition="'$(TargetFramework)' == 'net472'" Remove="$(MSBuildThisFileDirectory)..\App_Plugins\UmbracoCms.Integrations\Commerce\Shopify\RenderV9\*.*" />
```

ronaldbarendse (changes requested):
Besides aligning the view paths for the different versions, make sure to test whether the views are still copied over on .NET Framework (this doesn't seem to be the case by just looking at the code)...

src/Umbraco.Cms.Integrations.Commerce.Shopify/build/Umbraco.Cms.Integrations.Commerce.Shopify.targets
acoumb: Thanks for the suggestion @ronaldbarendse , updated accordingly. On V8 it works fine.

src/Umbraco.Cms.Integrations.Commerce.Shopify/Umbraco.Cms.Integrations.Commerce.Shopify.csproj
```diff
@@ -10,7 +10,7 @@
 		<PackageIconUrl></PackageIconUrl>
 		<PackageProjectUrl>https://github.com/umbraco/Umbraco.Cms.Integrations/blob/main/src/Umbraco.Cms.Integrations.Commerce.Shopify</PackageProjectUrl>
 		<RepositoryUrl>https://github.com/umbraco/Umbraco.Cms.Integrations</RepositoryUrl>
-		<Version>1.3.0</Version>
+		<Version>1.3.1</Version>
```
ronaldbarendse, line 13: Don't forget to update the version in the `package.manifest` and `package.xml` files!

src/Umbraco.Cms.Integrations.Crm.Dynamics/Umbraco.Cms.Integrations.Crm.Dynamics.csproj
```diff
@@ -10,7 +10,7 @@
 		<PackageIconUrl></PackageIconUrl>
 		<PackageProjectUrl>https://github.com/umbraco/Umbraco.Cms.Integrations/blob/main/src/Umbraco.Cms.Integrations.Crm.Dynamics</PackageProjectUrl>
 		<RepositoryUrl>https://github.com/umbraco/Umbraco.Cms.Integrations</RepositoryUrl>
-		<Version>1.3.6</Version>
+		<Version>1.3.7</Version>
```
ronaldbarendse, line 13: Same for this version 😄

ronaldbarendse (commented):
I've tested the Shopify integration on v13 and this correctly omits copying the `RenderV8` directory 👌🏻

It might be good the check all other build targets as well, because I did notice they are not aligning with the one provided in the package template for v13: https://github.com/umbraco/Umbraco-CMS/blob/v13/dev/templates/UmbracoPackage/buildTransitive/UmbracoPackage.targets.

PR #244, opened 2025-02-02
Fixed UmbracoTagsConverter not working

nickh364:
It would not work if tags storage format was set to csv


nickh364:
I know this could be done with a custom tags converter like in the Docs but it think by default it should work with both ways tags are formatted in Umbraco.

acoumb (approved):
Looks good to me.

PR #252, opened 2025-03-21
Fixed UmbracoMediaPickerConverter not working if value was Udi

nickh364:
This only worked before if the property value was JSON. Sometimes the value can be a Udi string.


acoumb:
Thank you for your contribution, the update will go out today.

PR #300, opened 2026-04-07
Added support for block list type when building objects to send to Algolia

jonathoncove1:
I thought this would be a useful addition to the integration, I'm open to suggestions!

I also hope I've setup the PR properly

### DIFF
--- a/src/Umbraco.Cms.Integrations.Search.Algolia/Converters/UmbracoBlockListConverter.cs
+++ b/src/Umbraco.Cms.Integrations.Search.Algolia/Converters/UmbracoBlockListConverter.cs
@@ -1,0 +1,77 @@
+﻿using Microsoft.Extensions.Logging;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Umbraco.Cms.Core.Models;
+using Umbraco.Cms.Core.PropertyEditors;
+using Umbraco.Cms.Core.PropertyEditors.ValueConverters;
+using Umbraco.Cms.Core.Serialization;
+using Umbraco.Cms.Integrations.Search.Algolia.Models;
+using Umbraco.Extensions;
+using static Umbraco.Cms.Core.Constants;
+
+namespace Umbraco.Cms.Integrations.Search.Algolia.Converters {    
+
+    public class UmbracoBlockListConverter : IAlgoliaIndexValueConverter {
+        private readonly IJsonSerializer _jsonSerializer;
+        private readonly BlockEditorConverter _blockEditorConverter;
+        private readonly ILogger<UmbracoBlockListConverter> _logger;
+        private const int charLimit = 250; 
+
+        public UmbracoBlockListConverter(IJsonSerializer jsonSerializer, BlockEditorConverter blockEditorConverter, ILogger<UmbracoBlockListConverter> logger) {
+            _jsonSerializer = jsonSerializer;
+            _blockEditorConverter = blockEditorConverter;
+            _logger = logger;
+        }
+
+        public string Name => PropertyEditors.Aliases.BlockList;
+
+        public object ParseIndexValues(IProperty property) {
+            var rawValue = property.GetValue();
+            List<string> stringValues = new List<string>();
+
+            if (rawValue == null || string.IsNullOrWhiteSpace(rawValue.ToString())) {
+                return "";
+            }
+
+            // Use Umbraco's built-in IJsonSerializer to deserialize the raw JSON
+            var blockListModel = _jsonSerializer.Deserialize<BlockListModelFromJson>(rawValue.ToString());
+
+            if (blockListModel == null) {
+                return "";
+            }
+
+            //go through the content blocks and try to get them as some useful type
+            foreach (var block in blockListModel.contentData) {
+
+                var content = _blockEditorConverter.ConvertToElement(block, PropertyCacheLevel.None, false);
+                if (content == null) continue;
+
+                //now we can go through the properties and convert their values, creating a big old list to return 
+                foreach(var prop in content.Properties) {
+                    bool isText = prop.PropertyType.DataType.Id == DataTypes.Textbox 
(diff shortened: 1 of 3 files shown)


src/Umbraco.Cms.Integrations.Search.Algolia/Converters/UmbracoBlockListConverter.cs
```diff
@@ -0,0 +1,77 @@
...
+using Umbraco.Extensions;
+using static Umbraco.Cms.Core.Constants;
+
+namespace Umbraco.Cms.Integrations.Search.Algolia.Converters {    
+
+    public class UmbracoBlockListConverter : IAlgoliaIndexValueConverter {
+        private readonly IJsonSerializer _jsonSerializer;
+        private readonly BlockEditorConverter _blockEditorConverter;
+        private readonly ILogger<UmbracoBlockListConverter> _logger;
+        private const int charLimit = 250;
```
rickbutterfield, line 21: I'm unsure about Algolia's functionality, but wouldn't a character limit this low cut most content pages off quite early?

src/Umbraco.Cms.Integrations.Search.Algolia/Converters/UmbracoBlockListConverter.cs
```diff
@@ -0,0 +1,77 @@
...
+                            stringValues.Add(toAdd);
+                        else 
+                            break;
+                        
+                    }
+                }
+
+            }
+
+            _logger.LogInformation("Converted block list to string: {stringValues}", stringValues);
```
rickbutterfield: This feels like it should be `Debug` rather than `Information` as it might get quite noisy in the logs

src/Umbraco.Cms.Integrations.Search.Algolia/Converters/UmbracoBlockListConverter.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        public string Name => PropertyEditors.Aliases.BlockList;
+
+        public object ParseIndexValues(IProperty property) {
+            var rawValue = property.GetValue();
+            List<string> stringValues = new List<string>();
+
+            if (rawValue == null || string.IsNullOrWhiteSpace(rawValue.ToString())) {
+                return "";
```
rickbutterfield: This should probably return the empty `stringValues`

src/Umbraco.Cms.Integrations.Search.Algolia/Converters/UmbracoBlockListConverter.cs
```diff
@@ -0,0 +1,77 @@
...
+            }
+
+            // Use Umbraco's built-in IJsonSerializer to deserialize the raw JSON
+            var blockListModel = _jsonSerializer.Deserialize<BlockListModelFromJson>(rawValue.ToString());
+
+            if (blockListModel == null) {
+                return "";
```
rickbutterfield: Should also return the empty `stringValues`

src/Umbraco.Cms.Integrations.Search.Algolia/Converters/UmbracoBlockListConverter.cs
```diff
@@ -0,0 +1,77 @@
...
+            foreach (var block in blockListModel.contentData) {
+
+                var content = _blockEditorConverter.ConvertToElement(block, PropertyCacheLevel.None, false);
+                if (content == null) continue;
+
+                //now we can go through the properties and convert their values, creating a big old list to return 
+                foreach(var prop in content.Properties) {
+                    bool isText = prop.PropertyType.DataType.Id == DataTypes.Textbox 
+                        || prop.PropertyType.DataType.Id == DataTypes.Textarea
+                        || prop.PropertyType.DataType.Id == DataTypes.RichtextEditor;
```
rickbutterfield: I think checking like this may be better...

```cs
var alias = prop.PropertyType.EditorAlias;
bool isText =
    alias == PropertyEditors.Aliases.TextBox
    || alias == PropertyEditors.Aliases.TextArea
    || alias == PropertyEditors.Aliases.RichText;
```

rickbutterfield (commented):
Hey @jonathoncove1! I've added a few comments from a quick review, hopefully nothing too major to resolve 😁

src/Umbraco.Cms.Integrations.Search.Algolia/Converters/UmbracoBlockListConverter.cs
jonathoncove1: The docs state it has a 10KB limit per record by default, but I was able to hit the limit with an average sized news article. The issue is that we can't check the overall size of the record being sent, so the limit is low to allow for other fields on the site getting us near the limit. There is the ability to split a record up into multiple, and link them, but that's outside the scope of this PR

src/Umbraco.Cms.Integrations.Search.Algolia/Converters/UmbracoBlockListConverter.cs
jonathoncove1: Fair enough, I've amended that

src/Umbraco.Cms.Integrations.Search.Algolia/Converters/UmbracoBlockListConverter.cs
jonathoncove1: Quite right. That makes the output more predictable, I've amended

src/Umbraco.Cms.Integrations.Search.Algolia/Converters/UmbracoBlockListConverter.cs
jonathoncove1: As above, amended

src/Umbraco.Cms.Integrations.Search.Algolia/Converters/UmbracoBlockListConverter.cs
jonathoncove1: That is a better way. I've updated with all the propertyeditors I could see that looked like purely text

PR #309, opened 2026-07-03
fix: hubspot form picker click selection and remove button overflow

Amalie-Wowern:
## Summary

- **Form picker click selection**: Clicking a form in the picker modal now selects it regardless of which part of the card is clicked (icon, name, or detail text). Root cause: `UUIRefNodeElement.handleOpenClick` calls `e.stopPropagation()` inside the shadow DOM, meaning `@click` on the host never fires. The correct event is `@open` (UUIRefEvent.OPEN), which UUI dispatches on the host element after the internal button is activated. Removing `selectable` ensures the `#open-part` button spans the full card width (with `selectable` it shrinks to icon-only size).
- **Remove button overflow**: The remove button no longer overflows the `uui-box` border. Fixed by removing the unnecessary `selectable` attribute from `uui-ref-node-form` in the property editor (form is already picked — selection UI is not needed there), and adding `:host { display: block }` so the element creates a block formatting context that contains the absolutely-positioned `uui-action-bar`.

## Test plan

- [ ] Add HubSpot API key (`Umbraco:Cms:Integrations:Crm:Hubspot:Settings:ApiKey`) and open the HubSpot form picker modal
- [ ] Click on the icon area (left side) of a form — should select and close the modal
- [ ] Click on the form name — should select and close the modal
- [ ] Click on the field detail text — should select and close the modal
- [ ] Filter forms by typing — filter should be case-insensitive
- [ ] After picking a form, verify the Remove button stays inside the `uui-box` border (no overflow)
- [ ] Click Remove — form should be cleared

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rickbutterfield (approved):
Looks great! Claude flagged that the same issue(s) exist in the ActiveCampaign and Dynamics implementations so I'll get matching PRs open for those

